PR #31, opened 2018-07-10
Include README in Python Package

ellismg:


mmdriley:
`setup.py` is automatically generated. Should this be done in cooperation with `pulumi-terraform`?

ellismg:
It is being done in cooperation with pulumi-terraform.  This change updates the dependency (note the pending change in the Gopkg.lock file).

This pulls in https://github.com/pulumi/pulumi-terraform/pull/197.

Or are you saying that pulumi-terraform should also generate the README.rst file in some fashion?  While that would be nice, a quick search did not indicate that there was an existing markdown to restructured text library for golang, so it would end up shelling out to pandoc at the end of the day regardless.

ellismg:
> Oh, frig, I missed that. Sorry. LGTM.

In your defense, we had conditioned you that changes to expect changes to `Gopkg.toml` as well, and now that we are doing the "right" thing, that invariant no longer holds.  Doesn't help that GitHub suppresses the diff for the lock file.

PR #36, opened 2018-07-28
Initial implementation of gcp.serverless.Function

mikhailshilkov:
Implemented gcp.serverless.Function in a way similar to aws.serverless.Function.

Limitations for now:
- HTTP-only, I haven't spend time on looking at other triggers yet
- Handler has no proper type definition. Might want to use express types? Should be adapted for other triggers too

Any feedback welcome!


sdk/nodejs/serverless/function.ts
```diff
@@ -0,0 +1,184 @@
+// Copyright 2016-2018, Pulumi Corporation.
```
lukehoban: Although a little roundabout - the way we add these "overlays" currently is via a mechanism in the `resources.go` that copies them into the generated folder instead of authoring directly in the generated folder.  See https://github.com/pulumi/pulumi-aws/blob/master/resources.go#L1831 for the equivalent in the AWS provider.

sdk/nodejs/serverless/function.ts
```diff
@@ -0,0 +1,184 @@
...
+import * as cloudfunctions from "../cloudfunctions";
+import * as storage from "../storage";
+
+import * as filepath from "path";
+import * as readPackageJson from "read-package-json";
+
+/**
+ * Handler is the signature for a serverless function.
+ */
+export type Handler = (request: any, response: any) => any;
```
lukehoban: Let's go ahead and use the types from `express` since Google Cloud does really provide live instances of those types - so really handy to get the rich typing automatically here.  See https://github.com/pulumi/examples/blob/master/gcp-ts-functions/gcpFunction.ts#L11.

sdk/nodejs/serverless/function.ts
```diff
@@ -0,0 +1,184 @@
...
+    public readonly options: FunctionOptions;
+    public readonly function: cloudfunctions.Function;
+    public readonly bucket: storage.Bucket;
+    public readonly bucketObject: storage.BucketObject;
+
+    constructor(name: string,
+        options: FunctionOptions,
+        func: Handler,
+        opts?: pulumi.ResourceOptions,
+        serialize?: (obj: any) => boolean) {
```
lukehoban: We shouldn't need this `serialize` any more in the AWS version of this function, and I'd suggest we leave it out here since there is no legacy requirement.

sdk/nodejs/serverless/function.ts
```diff
@@ -0,0 +1,184 @@
...
+        // serialized closure cannot reference.
+        serialize = serialize || (_ => true);
+        const finalSerialize = (o: any) => {
+            return serialize(o) && o !== this;
+        }
+
+        const handlerName = "handler";
+        const serializedFileNameNoExtension = "index";
+
+        let closure = pulumi.runtime.serializeFunction(func, {
```
lukehoban: If we were correctly linting this, we'd enforce that `const` be used when possible.  I think we don't have linting set up correctly here yet - but we should still use `const` consistently.

sdk/nodejs/serverless/function.ts
```diff
@@ -0,0 +1,184 @@
...
+                .reduce((obj, key) => {
+                    obj[key] = data.dependencies[key];
+                    return obj;
+                }, {});
+
+            const newPackageJson = {
+                name: data.name,
+                description: data.description,
+                dependencies: filteredDependencies
+            }
```
lukehoban: I don't actually know exactly what GCF does to install packages, but is it possible it needs more sections of the `package.json` to be fully correct?  Like `scripts` for install hooks?  Is it worth keeping everything by default and just overwriting `dependencies` field?

sdk/nodejs/serverless/index.ts
```diff
@@ -0,0 +1,2 @@
+// Export members:
+export * from "./function";
```
lukehoban: Trailing newline.

sdk/nodejs/serverless/function.ts
```diff
@@ -0,0 +1,184 @@
...
+     * Timeout (in seconds) for the function. Default value is 60 seconds. Cannot be more than 540 seconds.
+     */
+    readonly timeout?: pulumi.Input<number>;        
+   /**
+    * The packages relative to the program folder to not include the Function upload. This can be
+    * used to override the default serialization logic that includes all packages referenced by
+    * project.json (except @pulumi packages).  Default is `[]`.
+    */
+   excludePackages?: string[];
+}
```
lukehoban: We made several mistakes with `aws.serverless.Function` that ideally we wouldn't replicate here (and plan to fix in AWS).  Most notably, we don't really want to re-invent the set of options for a function here - we'd prefer to accept a `gcp.cloudfunctions.FunctionArgs` as input and merge it's options in directly with the source/entrypoint generated from the callback.  (Likely failing if conflicting settings are provided in the function args).

/cc @CyrusNajmabadi who is thinking about this for AWS and Azure as well.

lukehoban:
BTW - one question for @CyrusNajmabadi - should we put this in this package, or in a `gcp-serverless` package?  What's our plan for the corresponding functionality for AWS and Azure (I think we are inconsistent across those two currently?).

sdk/nodejs/serverless/function.ts
mikhailshilkov: Not sure I understand what you mean here... What are "overlays"?

sdk/nodejs/serverless/function.ts
mikhailshilkov: It's to you ofc, but sounds a bit leaky to me. We expose properties that can't be set (not just entrypoint, but e.g. `triggerHttp`). And we need an additional parameter `excludePackages` (AWS has more).
Serverless and raw are on different abstraction layers, so I'm not sure that options should be the same...

sdk/nodejs/serverless/function.ts
CyrusNajmabadi: @mikhailshilkov So, a good example of this is:

https://github.com/pulumi/pulumi-aws/blob/master/overlays/nodejs/ec2/instanceType.ts

The idea here is that *after* generating all the auto-generated code for our Resource types, we can add those 'overlay' files to the resultant folders to expose this extra API functionality through that package.   So, if you look up into https://github.com/pulumi/pulumi-aws/tree/master/sdk/nodejs/ec2, you'll see only that single file.  But if you then look at:

https://github.com/pulumi/pulumi-aws/tree/master/sdk/nodejs/ec2

You'll see the entire generated EC2 API *along with* that 'instanceType.ts' file.

--

So, the 'overlays' mechanism allows us to auto generate a large API, while also allowing us to 'inject' files into the final result.

sdk/nodejs/serverless/function.ts
CyrusNajmabadi: > It's to you ofc, but sounds a bit leaky to me.

'leaky' is one way to put it :)   But 'opinionated' is how i'd like ot put it.  Specifically, these APIs intentionally exist to create the real cloud resource.  i.e. in AWS we are trying to exactly expose the ability to create an AWS lambda.  In Azure we we are trying to exactly expose the abilty to create an Azure FunctionApp.  And the same will be true for GCP (i'm not knowledgable enough there to know precisely what that is).

Because we are helping you create those exact Resources, we want to give you the exact same level of control.  The purpose of these APIs is not really to abstract that away.  But simply to do the heavy lifting of allowing you to provide a javascript object that is then converted to the Cloud Lambda/FunctionApp/etc. iimplementation.

sdk/nodejs/serverless/function.ts
mikhailshilkov: I still don't get it , sorry. The last 2 links seem to be the same, and `instanceType.ts` seems the same in `overlay` and non-overlay folders...
What should I do?

joeduffy: It's a tad convoluted, so my apologies for the confusion.

Basically, in the AWS repo you have two files for the serverless equivalent:

* [`overlays/nodejs/serverless/function.ts`](https://github.com/pulumi/pulumi-aws/blob/master/overlays/nodejs/serverless/function.ts) - the file you write by hand
* [`sdk/nodejs/serverless/function.ts`](https://github.com/pulumi/pulumi-aws/blob/master/sdk/nodejs/serverless/function.ts) - the file that is generated when you `make`

Basically, the entire `sdk/*` directory is generated when you run `make`, mostly from Terraform metadata, however it also mixes in directives from the [`resources.go`](https://github.com/pulumi/pulumi-aws/blob/master/resources.go) file. In this specific case, there is [a directive in that file](https://github.com/pulumi/pulumi-aws/blob/4c982c2f275c6b39383acac761f44449c099b461/resources.go#L1831) which is responsible for merging that file in `overlays/...` into the resulting built `sdk/...` directory. As a result, everything underneath `sdk/*` is generated.

I do get the benefit of being consistent here, although honestly, @lukehoban, @CyrusNajmabadi, we may want to revisit this now that we commit the generated sdk directories...partly, we did it this way originally because we didn't commit them, although there are still possibly reasons (e.g., it's clearer what is generated versus not).

At the end of the day, I would *love* to land this change, so I'm happy to help refactor it to conform to this alternative model, provided all is working and otherwise well.

**Thank you** for doing this! 🎉

mikhailshilkov: Ah, okay, I'll move it to `overlays` folder.

mikhailshilkov: Wait, there's no `overlays` folder in GCP. I'm confused again about what needs to happen.

sdk/nodejs/serverless/function.ts
mikhailshilkov: Do we still want `excludePackages` property? Do something like
``` typescript
export interface FunctionOptions extends cloudfunctions.FunctionArgs {   
   excludePackages?: string[];
}
```
or do we want to accept exactly `FunctionArgs`?

CyrusNajmabadi: @mikhailshilkov I think it's reasonable to do that (and, in fact, we should move closer to that in aws-serverless).  We do want to allow users to pass all (or nearly all) of the options to configure the function.  On top of that, we would want things like excludePackages to control the parts of logic that Pulumi adds itself (like packing up the code).  So i would go with this latest suggestion so that you're not blocked here.  We may end up tweaking this a bit as we solidify aws-serverless and whatnot.

mikhailshilkov: Hmm... This turns out problematic because the original `FunctionArgs` have parameters like `sourceArchiveBucket` mandatory, while `FunctionOptions` should actually depend *not* to have them... I guess I can't declare a type based on the other type, but *excluding* given properties?

CyrusNajmabadi: You *technically* can, using some advanced typescript features.  You can see one of the ytpes that can handle that here: https://github.com/pulumi/pulumi-aws-serverless/pull/31/files#diff-7d9b0842656b84e670d724366d103be7

Effectively, you could say: 

```ts
type FunctionArgs = Omit<cloudfunctions.FunctionArgs, "sourceArchiveBucket" | "otherPropName" | "etc">
```

However, i'm hesitant about thsi approach because it can be hard to comprehend.  i.e. it helps TS strongly check that your types have hte right shape.  But it can then be harder for someone to grok waht is actually going on.

CyrusNajmabadi:
> BTW - one question for @CyrusNajmabadi - should we put this in this package, or in a gcp-serverless package? What's our plan for the corresponding functionality for AWS and Azure (I think we are inconsistent across those two currently?).

I don't think we have a plan per-se.  It's worth deciding now since any changes in the future would be more breaking changes.  I think my overall preference would be:

1. Keep the core cloud packages very simple.  So `pulumi/aws`/`pulumi/azure`/`pulumi/gcp` are effectively just the projections of those cloud provides into Resources that users can use, without much special logic in them at all.
2. Logic around bridging hte gap between JavaScript functions and Cloud Lambdas/FunctionApps/etc. can go in `pulumi/***-serverless`.  so `pulumi/aws-serverless`/`pulumi/azure-serverless`/`pulumi/gcp-serverless`.  
3. Logic to help easily connect Resources and expose things like cloud-triggers also stays in those same packages, and should feel very much like it it fits the patterns and terminologies of those respective Cloud providers.  So, for example, `serverless.s3.onObjectCreated(bucket, ...)` (in the AWS lib), or `serverless.storage.onBlobEvent(account, ...)`, etc. etc.

Effectively, keep the core libraries simple without much actual custom logic.  Put all the logic for lambdas/functions/triggers/connections in one package.  

Does that seem reasonable to use @lukehoban ?

joeduffy:
@mikhailshilkov I opened https://github.com/mikhailshilkov/pulumi-gcp/pull/1 to show what using overlays would look like. I think that if we get that in, we're good to merge on this.

The conversation @lukehoban and @CyrusNajmabadi are having at the end of this PR is a longer-term discussion we need to have (and frankly looks like we're leaning towards doing precisely what your PR does). Either way, shouldn't hold up getting this very good work landed so people can use GCP functions more easily!

lukehoban:
> Either way, shouldn't hold up getting this very good work landed so people can use GCP functions more easily!

Agreed - let's get this in once we've integrated the overlay changes, and then we can tackle the factoring as part of reviewing this factoring across AWS, Azure and GCP.

joeduffy:
LGTM! Seems there are conflicts, if you want to resolve them, happy to merge as soon as you're ready! We finish our major release today so looks like this will be part of it 😁🎉

PR #43, opened 2018-09-06
Update pulumi-terraform.

pgavlin:
This brings in support for default config and property values provided
by environment variables.


lukehoban:
Do we expect to have changes here to pick up default `project`, `zone`, etc. from environment variables?

pgavlin:
> Do we expect to have changes here to pick up default project, zone, etc. from environment variables?

Yes. I need to update `resources.go` as well.

PR #53, opened 2018-10-16
serverless.Function: support triggerBucket and triggerTopic

### LINKED ISSUE #52: serverless.Function: add support for pubsub trigger
Currenlty `serverless.Function` only supports http as trigger apparently.
It'd like to use a pubsub topic as trigger.
This should be easily possible by passing`triggerTopic` here: https://github.com/pulumi/pulumi-gcp/blob/master/overlays/nodejs/serverless/function.ts#L114

geekflyer:
fixes #52

### DIFF
--- a/overlays/nodejs/serverless/function.ts
+++ b/overlays/nodejs/serverless/function.ts
@@ -56,6 +56,16 @@ export interface FunctionOptions {
      * Timeout (in seconds) for the function. Default value is 60 seconds. Cannot be more than 540 seconds.
      */
     readonly timeout?: pulumi.Input<number>;
+    /**
+     * Google Cloud Storage bucket name. Every change in files in this bucket will trigger function execution. Cannot be used with triggerTropic.
+     * If neither triggerTopic nor triggerBucket is set an http triggered function will be created instead.
+     */
+    readonly triggerBucket?: pulumi.Input<string>;
+    /**
+     * Name of a Pub/Sub topic. Every message published in this topic will trigger function execution with message contents passed as input data. Cannot be used with triggerBucket.
+     * If neither triggerTopic nor triggerBucket is set an http triggered function will be created instead.
+     */
+    readonly triggerTopic?: pulumi.Input<string>;
    /**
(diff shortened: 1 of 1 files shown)


overlays/nodejs/serverless/function.ts
```diff
@@ -111,14 +121,16 @@ export class Function extends pulumi.ComponentResource {
             entryPoint: handlerName,
             sourceArchiveBucket: this.bucket.name,
             sourceArchiveObject: this.bucketObject.name,
-            triggerHttp: true,
+            triggerHttp: !(options.triggerBucket || options.triggerTopic),
```
lukehoban: These values are `Input<T>` so you would likely need to do `pulumi.all([options.triggerBucket, options.triggerTopic]).apply(([triggerBucket, triggerTopic]) => !(options.triggerBucket || options.triggerTopic))`.   I suspect as written this will not work correctly.

overlays/nodejs/serverless/function.ts
```diff
@@ -111,14 +121,16 @@ export class Function extends pulumi.ComponentResource {
...
 
             availableMemoryMb: options.availableMemoryMb,
             description: options.description,
             labels: options.labels,
             project: options.project,
             region: options.region,
-            timeout: options.timeout
+            timeout: options.timeout,
+            triggerBucket: options.triggerBucket,
+            triggerTopic: options.triggerTopic
```
lukehoban: In the latest TF Google provider, these properties have actually been deprecated.  We haven't pulled that in yet - but we will very soon.  Given that, I suspect we will want to avoid building on top of them here.  https://github.com/terraform-providers/terraform-provider-google/blob/master/CHANGELOG.md#1190-october-08-2018

overlays/nodejs/serverless/function.ts
lukehoban: But actually also not sure this is what we want.  Should we just expose `triggerHttp` as a property like in the TF provider and pass it through here.  That would be a breaking change (it would not be `true` by default) - but I think it's more likely the right behaviour here to align with the underlying provider.

PR #54, opened 2018-10-23
Update to the latest TF provider.

pgavlin:
These changes update to v1.19.1 of the TF GCP provider.


jen20:
This looks good to me - a local build doesn't show any unmapped resources. I'm not sure the assumptions made about doc layout still holds, since I _do_ see a ton of "warning: Expected only a single H1 in markdown for resource".

lukehoban:
> “warning: Expected only a single H1 in markdown for resource".

Yes - these have always been there for th GCP provider. We need to update the doc extraction heuristics to better handle the different structure of the GCP docs.

PR #60, opened 2018-11-07
Pull in updates to pulumi/pulumi-terraform

jen20:
This commit modifies the constraints in Gopkg.toml to allow solving for an update of `pulumi-terraform`, given the changes to the dependencies of `pulumi-terraform`.

This pulls in the changes to use Terraform importers.

I have tested the examples with a build using this set of dependencies, and there are no related failures (some of the instance startup scripts in the examples don't produce the desired results with the current generation of images, but that is unrelated to these changes).

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -149,6 +149,15 @@
   version = "v1.13.1"
 
 [[projects]]
+  name = "github.com/dustin/go-humanize"
+  packages = [
+    ".",
+    "english"
+  ]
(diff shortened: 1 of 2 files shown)


jen20:
I was somewhat nervous about that too - with that constraint, `dep` 0.4.1 would not solve, despite the fact that it picks that revision anyway. Hopefully for the next revision we can constrain this again if necessary.

PR #63, opened 2018-11-12
Update pulumi-terraform reference

ellismg:


ellismg:
> This looks like it's missing a new generated file tables.py, did it get generated when you ran tfgen locally?

Good catch.  It was missing.  Added it to the PR and repushed.

PR #74, opened 2018-12-29
python: open README.rst with utf-8 encoding

vrutkovs:
This would ensure pulumi-gcp can be installed using python3-pip


stack72:
Closing this as this has been fixed in our new upstream pulumi-terraform-bridge project

PR #75, opened 2018-12-29
Ingest latest Terraform code-gen

joeduffy:
This ingests the latest Terraform code-gen, bringing in the
ability to convert HCL examples to Pulumi in the API docs.

Notes on metrics:

* 242 docs sections ignored
* 19 code blocks failed to convert
* 160 code blocks successfully converted

### DIFF
--- a/sdk/go/gcp/organizations/getFolder.go
+++ b/sdk/go/gcp/organizations/getFolder.go
@@ -10,6 +10,26 @@ import (
 // Use this data source to get information about a Google Cloud Folder.
 // 
 // ```hcl
+// # Get folder by id
+// data "google_folder" "my_folder_1" {
+//   folder = "folders/12345"
+//   lookup_organization = true
+// }
+// 
+// # Search by fields
+// data "google_folder" "my_folder_2" {
+//   folder = "folders/23456"
+// }
+// 
+// output "my_folder_1_organization" {
+//   value = "${data.google_folder.my_folder_1.organization}"
+// }
+// 
+// output "my_folder_2_parent" {
+//   value = "${data.google_folder.my_folder_2.parent}"
+// }
+// 
+// ```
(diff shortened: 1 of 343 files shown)


lukehoban:
I think this may be related to https://github.com/pulumi/pulumi-terraform/pull/288. I’m not positive we ever regenerated the GCP provider after those changes?

sdk/nodejs/compute/address.ts
```diff
@@ -4,6 +4,89 @@
...
+ * ## Example Usage
+ * 
+ * ```typescript
+ * import * as pulumi from "@pulumi/pulumi";
+ * import * as gcp from "@pulumi/gcp";
+ * 
+ * const google_compute_address_ip_address = new gcp.compute.Address("ip_address", {
+ *     name: "my-address",
+ * });
+ * ```
```
joeduffy: This is funky, not sure what is going on here. Need to investigate.

sdk/go/gcp/organizations/getFolder.go
```diff
@@ -10,6 +10,26 @@ import (
 // Use this data source to get information about a Google Cloud Folder.
 // 
 // ```hcl
+// # Get folder by id
```
joeduffy: Looks like this is a code snippet embedded inside of the resource's top level comment, not within an example section.

joeduffy:
Ahh, ok, makes sense. That fix on its own didn't seem to do the trick for me, as it lead to a big header at the top of many resources with the front-matter `# WARNING THIS IS CODE GENERATED`, etc, etc, rather than it finding the resource H1 as intended. My PR over there seems like a step forward.

jen20:
This is likely superseded by #76, since it pulls in the latest versions of all dependencies and has the output of `tfgen` committed as part of it.

PR #76, opened 2019-01-07
Update to upstream provider v1.20.0, and introduce govendor-override

jen20:
This pull request updates the Google provider to v1.20.0 of the upstream provider, and changes the method by which dependencies are resolved to use `govendor-override`, bringing the Google provider in line with the other Terraform-backed providers.


pgavlin:
Was the `Gopkg.lock` file intentionally deleted?

jen20:
No, something is missing there, I'll take a look.

jen20:
This should be fixed, though (sadly) at the cost of the commit history here, though I'm not overly sure that matters too much for individual new resources.

jen20:
Worth noting that there is no documentation as part of the upstream provider for `google_compute_router_nat` - although it did appear after the release. I'm loathe to maintain a fork just to cherry-pick one file, so have left that (new) resource without documentation for now.

PR #82, opened 2019-01-18
Base pulumi-gcp on the google-beta Terraform provider

### LINKED ISSUE #57: consider adding support for terraform provider google beta
It came recently to my attention that google is apparently planning to branch very soon (2.0.0 release) the google terraform provider into 2 flavors (https://www.terraform.io/docs/providers/google/provider_versions.html). One regular (stable) and one beta provider. The difference between the two is that the regular one only supports GCP APIs which are marked as stable whereas the beta one also supports beta APIs and features.
Singe google has a philosophy of keeping things pretty long in beta, the 1.x.x provider contains a bunch of beta features that are already used by customers in production (Personally we use gke nodepool taints, which are considered beta by google and will be removed in the future provider version).

While this is currently not an issue (since pulumi adopts the 1.x.x version of the terraform provider) I think this may be come an issue in near future and I wanted to bring this to pulumi's attention early enough. I'm not sure what's the right solution but it certainly would be good if pulumi customers have the choice to only use stable or beta google apis, just like they would have when using terraform. This may require creating 2 pulumi gcp flavors as well.

On another note: Google recently moved to auto-generate most of the google terraform provider via some meta module called 'magic-modules': https://github.com/GoogleCloudPlatform/magic-modules/ . The idea of magic-modules is in my understanding to auto-generate SDKs for different languages and platforms (currently terraform, ansible, puppet, chef). Mid term this might be an opportunity for pulumi to auto-generate the pulumi gcp module directly based on the magic-modules instead of the terraform provider.

jen20:
Fixes #57.

The beta provider is described [here](https://www.terraform.io/docs/providers/google/provider_versions.html) as being interchangeable with the regular provider, except with additional support for beta properties and resources.

This also updates `pulumi-terraform`, part of pulumi/pulumi-terraform#306.


jen20:
This isn't actually sufficient to move the provider over - there's more to be done.

PR #83, opened 2019-01-19
Update pulumi-terraform to b0770d3

jen20:
This PR updates pulumi-terraform to b0770d3 which pulls in the new doc generation from @joeduffy, the python updates from @swgillespie and the update to the Pulumi dependency from @ellismg.

It's possible the Makefile will need modification in order to install the correct providers for documentation - I'll wait on CI to see if that needs doing.


Makefile
```diff
@@ -39,6 +37,16 @@ build::
...
 		cd ./bin && $(PYTHON) setup.py build sdist
 
+provider::
+	go install -ldflags "-X github.com/pulumi/pulumi-gcp/pkg/version.Version=${VERSION}" ${PROJECT}/cmd/${PROVIDER}
+
+tfgen::
+	go install -ldflags "-X github.com/pulumi/pulumi-gcp/pkg/version.Version=${VERSION}" ${PROJECT}/cmd/${TFGEN}
+
+install_plugins::
+	curl -fsSL https://get.pulumi.com | sh
```
joeduffy: Should we just require that `pulumi` is on the path, rather than auto-installing from `get.pulumi.com`?

jen20: Long term I think that's a better plan (or alternatively, we might want to make tf2pulumi learn to install providers so that having pulumi becomes unnecessary)

lukehoban:
For this, and the other similar provider updates today, should we update `CHANGELOG` with details of new TF provider version and a note about examples in docs?

jen20:
@lukehoban Yes, I'm intending to push out new releases of everything once CI on master has been verified.

Makefile
jen20: I think the approach in cf4bb76 is a happy medium for now - it won't result in surprising results on developer machines, but will install in CI without making modifications to the scripts for now. Once we figure out a better story for plugins here I'll change the approach.

PR #84, opened 2019-01-20
Update Gopkg.lock for dep 0.5

jen20:
This PR updates Gopkg.lock for the version generated by dep 0.5.


jen20:
I've just rebased and force pushed this to ensure that no changes on master have had an impact.

PR #101, opened 2019-03-04
Retype gcp.storage.BucketObject::source as Asset

jen20:
This is a fix for the bug seen in pulumi-gcp#87 - I suspect there will be fallout for examples and so forth, hence opening the PR early to check CI.


jen20:
The build here should be fixed if rebased on #106 to pick up new `pulumi-terraform` changes.

PR #106, opened 2019-03-08
Upgrade to the Google Beta Provider v2.1.0

jen20:
This pull request upgrades the provider to be based on v2.1.0 of the Google Beta Terraform provider. This contains a large number of breaking changes, so relies on the work done to align the SDK and provider versions.

Additionally, it uses the updated `gomod-override` from pulumi/scripts#47, since upstream has switched to Go modules.

Fixes #86.


jen20:
Closed in favour of #110.

PR #110, opened 2019-03-19
Update to the Google Beta Provider v2.2.0

### LINKED ISSUE #57: consider adding support for terraform provider google beta

jen20:
This pull request supersedes #106, which was difficult to rebase, and upgrades to the latest v2.2.0 of the beta provider now we are ready to make breaking changes.

From upstream, a list of breaking changes (some of which may need addressing in examples, pending the result of CI):

- **bigtable**: google_bigtable_instance zone field is no longer inferred from the provider.
- **bigtable**: google_bigtable_table now reads family from the table's column family in Cloud Bigtable instead of creating a new column family
- **bigtable**: google_bigtable_instance.cluster.num_nodes will fail at plan time if DEVELOPMENT instances have num_nodes = "0" set explicitly. If it has been set, unset the field.
- **cloudbuild**: google_cloudbuild_trigger.build.step.args is now a list instead of space separated strings.
- **cloudfunctions**: google_cloudfunctions_function.retry_on_failure has been removed. Use event_trigger.failure_policy.retry instead.
- **cloudfunctions**: google_cloudfunctions_function.trigger_bucket and google_cloudfunctions_function.trigger_topic have been removed. Use event trigger instead.
- **composer**: google_composer_environment.node_config.zone is now Required.
- **compute**: google_compute_instance, google_compute_instance_from_template metadata field is now authoritative and will remove values not explicitly set in config.
- **compute**: google_compute_region_instance_group_manager field update_strategy is now deprecated in the beta provider only. It will only function in the google provider,
- **compute**: google_compute_global_forwarding_rule field labels is now removed
- **compute**: google_compute_project_metadata resource is now authoritative and will remove values not explicitly set in config.
- **compute**: google_compute_url_map resource is now authoritative and will remove values not explicitly set in config.
- **compute**: google_compute_snapshot.snapshot_encryption_key_raw, google_compute_snapshot.snapshot_encryption_key_sha256, google_compute_snapshot.source_disk_encryption_key_raw, google_compute_snapshot.source_disk_encryption_key_sha256 fields are now removed. Use google_compute_snapshot.snapshot_encryption_key.0.raw_key, google_compute_snapshot.snapshot_encryption_key.0.sha256, google_compute_snapshot.source_disk_encryption_key.0.raw_key, google_compute_snapshot.source_disk_encryption_key.0.sha256 instead.
- **compute**: google_compute_instance_group_manager is no longer imported by the provider-level region. Set the appropriate provider-level zone instead.
- **compute**: google_compute_region_instance_group_manager.update_strategy in the google-beta provider has been removed.
- **compute**: google_compute_instance, google_compute_instance_template, google_compute_instance_from_template have had the network_interface.address field removed.
- **compute**: google_compute_instance has had the network_interface.access_config.assigned_nat_ip field removed
- **compute**: google_compute_disk is no longer imported by the provider-level region. Set the appropriate provider-level zone instead.
- **compute**: google_compute_router_nat.subnetwork.source_ip_ranges_to_nat is now Required inside subnetwork blocks.
- **compute**: google_compute_ssl_certificate's private_key field is no longer stored in state in cleartext; it is now SHA256 encoded.
- **container**: google_container_cluster fields (private_cluster, master_ipv4_cidr_block) are removed. Use private_cluster_config and private_cluster_config.master_ipv4_cidr_block instead.
- **container**: google_container_node_pool's name_prefix field has been restored and is no longer deprecated.
- **sql**: google_sql_database_instance resource is now authoritative and will remove values not explicitly set in config.
- **endpoints**: google_endpoints_service.protoc_output was removed. Use google_endpoints_service.protoc_output_base64 instead.
- **resourcemanager**: google_project_iam_policy is now authoritative and will remove values not explicitly set in config. Several fields were removed that made it authoritative: authoritative, restore_policy, and disable_project. This resource is very dangerous! Ensure you are not using the removed fields (authoritative, restore_policy, disable_project).
- **resourcemanager**: Datasource google_service_account_key.service_account_id has been removed. Use the name field instead.
- **resourcemanager**: google_project.app_engine has been removed. Use the google_app_engine_application resource instead.
- **resourcemanager**: google_organization_custom_role.deleted is now an output-only attribute. 
- **resourcemanager**: google_project_custom_role.deleted is now an output-only attribute. 
- **serviceusage**: google_project_service will now error instead of silently disabling dependent services if disable_dependent_services is unset.
- **storage**: google_storage_object_acl.role_entity is now authoritative and will remove values not explicitly set in config. Use google_storage_object_access_control for fine-grained management.
- **storage**: google_storage_default_object_acl.role_entity is now authoritative and will remove values not explicitly set in config.
- **iam**: google_*_iam_binding Change all IAM bindings to be authoritative

Fixes #57.


resources.go
```diff
@@ -509,8 +542,16 @@ func Provider() tfbridge.ProviderInfo {
 			"google_runtimeconfig_config":   {Tok: gcpResource(gcpRuntimeConfig, "Config")},
 			"google_runtimeconfig_variable": {Tok: gcpResource(gcpRuntimeConfig, "Variavble")},
```
swgillespie: Not your code, but can you fix this typo while you're at it:
```suggestion
			"google_runtimeconfig_variable": {Tok: gcpResource(gcpRuntimeConfig, "Variable")},
```

Technically a breaking change, so might as well do it while we're at it?

resources.go
```diff
@@ -373,6 +404,7 @@ func Provider() tfbridge.ProviderInfo {
 			"google_compute_instance_group_manager":        {Tok: gcpResource(gcpCompute, "InstanceGroupManager")},
 			"google_compute_instance_template":             {Tok: gcpResource(gcpCompute, "InstanceTemplate")},
 			"google_compute_interconnect_attachment":       {Tok: gcpResource(gcpCompute, "InterconnectAttachment")},
+			"google_compute_managed_ssl_certificate":       {Tok: gcpResource(gcpCompute, "MangedSslCertificate")},
```
swgillespie: We use uppercase SSL elsewhere, like `SSLCertificate` and `SSLPolicy` - how about `ManagedSSLCertificate`?

resources.go
```diff
@@ -668,6 +713,12 @@ func Provider() tfbridge.ProviderInfo {
 					Source: "datasource_client_config.html.markdown",
 				},
 			},
+			"google_client_openid_userinfo": {
+				Tok: gcpDataSource(gcpOrganization, "getClientOpenIdUserInfo"),
```
swgillespie: Similar naming convention here re: SSL: `getClientOpenIDUserInfo`

resources.go
lukehoban: I think unfortunately we've been really inconsistent on this.  In the AWS provider, my sense is we've nearly always used `Ssl` for three letter acronyms.  Would love to do an audit on this for our next round of breaking changes.

resources.go
```diff
@@ -842,6 +893,18 @@ func Provider() tfbridge.ProviderInfo {
...
+			"google_kms_crypto_key": {
+				Tok: gcpDataSource(gcpKMS, "getKMSCryptoKey"),
+				Docs: &tfbridge.DocInfo{
+					Source: "google_kms_crypto_key.html.markdown",
+				},
+			},
+			"google_kms_key_ring": {
+				Tok: gcpDataSource(gcpKMS, "getKMSKeyRing"),
+				Docs: &tfbridge.DocInfo{
+					Source: "google_kms_key_ring.html.markdown",
```
lukehoban: If there are patterns here that are consistent but not being picked up automatically - we may want to embed theose patterns in the search heuristics used in `pulumi-terraform` so that we don't need as many manual annotations like this.  I'm not sure why this particular one doesn't get picked up automatically with the latest `pulumi-terraform` - I thought we added a rule to handle exactly this pattern.

jen20: Not sure in this particular case, but I agree it's worth auditing what is being used and adding the various patterns where possible.

resources.go
jen20: After discussion with @lukehoban it sounds like we'll go for these next time around after we decide a project-wide policy.

jen20:
I've run all of the examples in `pulumi/examples` with this upgraded version of the provider, and all seems well - I think we are good to release a v0.18.0 here.

kurlzor:
Is this supposed to be working already ? I've updated the package using npm but when I try to create a private DNS `ManagedZone` on Google Cloud Platform, I get this:
```
error: gcp:dns/managedZone:ManagedZone resource 'typhon' has a problem: : invalid or unknown key: visibility
error: gcp:dns/managedZone:ManagedZone resource 'typhon' has a problem: : invalid or unknown key: private_visibility_config
```

PR #119, opened 2019-04-01
Update to upstream v2.3.0

jen20:
Update to v2.3.0 of the Google Beta Terraform Provider.


sdk/nodejs/dns/policy.ts
```diff
@@ -22,41 +22,6 @@ import * as utilities from "../utilities";
...
- *     networks: [
- *         {
- *             networkUrl: network_1.selfLink,
- *         },
- *         {
- *             networkUrl: network_2.selfLink,
- *         },
- *     ],
- * });
- * ```
```
swgillespie: Do you know why this happens? I see this sometimes in my own diffs (losing tf2pulumi examples)

jen20: Usually some kind of error has been introduced into the HCL when we lose an example like this - I'll open a new issue to track them down and see about recovering them.

Issue #93, opened 2019-02-11
Override "Managed by Terraform" defaults


clstokes:
GCP managed zones have a default description of `Managed by Terraform`. We should override this to `Managed by Pulumi` as we do in [AWS](https://github.com/pulumi/pulumi-aws/blob/9c1de153c8681172a53c63244daab25e7636b152/resources.go#L204).

Occurrences: 
- https://github.com/terraform-providers/terraform-provider-google/blob/6ee10e3bf3713085d5571b9190451769b21fbe89/google/resource_dns_managed_zone.go#L57

PR #134, opened 2019-04-08
Add ability to override a managed by description

stack72:
Fixes: #93

This introduces the work to allow a resource to override a "Managed
by Terraform" description

### DIFF
--- a/resources.go
+++ b/resources.go
@@ -81,6 +81,9 @@ func gcpResource(mod string, res string) tokens.Type {
 	return gcpType(mod+"/"+fn, res)
 }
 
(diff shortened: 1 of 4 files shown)


jen20:
This build is failing because of https://github.com/pulumi/pulumi-gcp/blob/cc6c26e1a0184600034fb97c1f9e03b258c4baea/.travis.yml#L21 - in a fork, the secret is unavailable. We need to modify the build script here to test whether we're in a fork or not before attempting, and just skip anything requiring credentials if so.

joeduffy:
This is awesome, LGTM :+1:, thanks @stack72.

PR #139, opened 2019-04-26
Updating to terraform-provider-google-beta v.2.5.0

lukehoban:
Bring https://github.com/pulumi/pulumi-gcp/pull/137 over to a non-fork for testing.

PR #137, opened 2019-04-22
Updating to Terraform provider v.2.5.0

stack72:

PR #140, opened 2019-04-28
Update to provider v2.5.0

jen20:
Supersedes #139 and #137.


jen20:
Looks like this is failing in a very similar way to how @lukehoban saw. I wonder if the test framework has added new assertions that have been pulled in somehow. I'll see what I can reproduce locally.

jen20:
Updating `pulumi-terraform` and pulling in `v2.5.1` has fixed this.

PR #154, opened 2019-05-29
Fix CI for GCP

jen20:
This pull request updates the credentials for GCP to allow CI to move to a new project such that leaked resources can be more easily identified.


jen20:
The project setup is currently not quite correct, hence the build failures here. I'll kick this build off again once this is rectified.

PR #180, opened 2019-07-16
Upgrade to v2.11.0 of the Google Beta Terraform Provider

stack72:


pgavlin:
Looks like we're missing a couple of resources: https://travis-ci.com/pulumi/pulumi-gcp/builds/119389427#L1593-L1594

stack72:
> Looks like we're missing a couple of resources: [travis-ci.com/pulumi/pulumi-gcp/builds/119389427#L1593-L1594](https://travis-ci.com/pulumi/pulumi-gcp/builds/119389427#L1593-L1594)

Funny you found those! These are actually resources we wrote that don't exist in Terraform! There are also some docs we need to sort out - this is something I need to raise upstream though...

PR #188, opened 2019-07-30
Reenable the Loadbalancer test

stack72:
Fixes: #183

After https://github.com/GoogleCloudPlatform/magic-modules/pull/2084 was merged
we can see that we should be using an actual IP address rather than the
self link due to how the resolution appears. This means that the functionality
has changed and therefore, we will update the example

### DIFF
--- a/examples/examples_test.go
+++ b/examples/examples_test.go
@@ -100,14 +100,12 @@ func TestExamples(t *testing.T) {
 	}
 
 	longTests := []integration.ProgramTestOptions{
-		//jsBase.With(integration.ProgramTestOptions{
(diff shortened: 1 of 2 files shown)


stack72:
FYI, this is in the magic-modules docs:

```
The server accepts a literal IP address or a URL reference to an existing
          Address resource. The following examples are all valid but only the first
          will prevent a permadiff. If you are using `google_compute_address` or
          similar, interpolate using `.address` instead of `.self_link` or similar
          to prevent a diff on re-apply.
```

stack72:
Self merging as this is only a test renabling again

PR #189, opened 2019-07-30
Upgrade to pulumi-terraform 6bcafb609d

stack72:


sdk/nodejs/composer/getImageVersions.ts
```diff
@@ -28,6 +28,13 @@ import * as utilities from "../utilities";
  */
 export function getImageVersions(args?: GetImageVersionsArgs, opts?: pulumi.InvokeOptions): Promise<GetImageVersionsResult> & GetImageVersionsResult {
     args = args || {};
+    if (!opts) {
+        opts = {}
+    }
+
+    if (!opts.version) {
+        opts.version = utilities.getVersion();
+    }
```
lukehoban: Not that it matters much since this is generated code - but seeing this in context:
1. It’s strange that we use two different styles for defaulting args and opts respectively. 
2. It’s strange that we have an empty line where we do.

PR #194, opened 2019-08-08
Adding python tests for resource and datasource

stack72:
This helps to test any backwards incompatibilities in our code


stack72:
Self merging this - this only adds more tests for Python to GCP

PR #199, opened 2019-08-20
Add the work to allow tests to run upgrades of providers

stack72:
Related to: pulumi/pulumi#2757


examples/examples_test.go
```diff
@@ -74,7 +74,8 @@ func TestExamples(t *testing.T) {
 	shortTests := []integration.ProgramTestOptions{
 		jsBase.With(integration.ProgramTestOptions{Dir: path.Join(cwd, "minimal")}),
 		jsBase.With(integration.ProgramTestOptions{
-			Dir: path.Join(cwd, "serverless"),
+			Dir:           path.Join(cwd, "serverless"),
+			RunUpdateTest: true,
```
ellismg: Similar feedback here, would be nice to move RunUpdateTest to the `jsBase` thing.

stack72: Same response as with AWS - will test and come back to make sure we do this if it works

PR #216, opened 2019-09-10
Update to v2.14.0 of the Google Beta Terraform Provider

### LINKED ISSUE #212: panic: fatal: An assertion has failed: Nested type "RegistryEventNotificationConfig" already exists with a different declaration.
When upgrading to [v2.12.0](https://github.com/terraform-providers/terraform-provider-google-beta/releases/tag/v2.12.0) of the TF provider, @stack72 ran into this assert:

```
panic: fatal: An assertion has failed: Nested type "RegistryEventNotificationConfig" already exists with a different declaration.
```

The root cause is due to: https://github.com/terraform-providers/terraform-provider-google-beta/commit/f082bba6737c14c6736a5d93ffba998ff41e1f66

The `google_cloudiot_registry` resource (which we project as `gcp.kms.Registry`) was modified as follows:

The existing `event_notification_config` (TypeMap) property _singular_ is now deprecated in favor of a new property named `event_notification_configs` (TypeList) _plural_.

```go
"event_notification_config": {
	Type:          schema.TypeMap,
	Optional:      true,
	Computed:      true,
	Deprecated:    "eventNotificationConfig has been deprecated in favor of eventNotificationConfigs (plural). Please switch to using the plural field.",
	ConflictsWith: []string{"event_notification_configs"},
	Elem: &schema.Resource{
		Schema: map[string]*schema.Schema{
			"pubsub_topic_name": {
				Type:             schema.TypeString,
				Required:         true,
				DiffSuppressFunc: compareSelfLinkOrResourceName,
			},
		},
	},
},
"event_notification_configs": {
	Type:          schema.TypeList,
	Optional:      true,
	Computed:      true,
	MaxItems:      10,
	ConflictsWith: []string{"event_notification_config"},
	Elem: &schema.Resource{
		Schema: map[string]*schema.Schema{
			"pubsub_topic_name": {
				Type:             schema.TypeString,
				Required:         true,
				DiffSuppressFunc: compareSelfLinkOrResourceName,
			},
			"subfolder_matches": {
				Type:         schema.TypeString,
				Optional:     true,
				ValidateFunc: validateCloudIotRegistrySubfolderMatch,
			},
		},
	},
},
```

The problem is we first generate the nested type for the first (deprecated) property as:

```typescript
interface RegistryEventNotificationConfig {
    pubsubTopicName: string;
}
```

We then try to generate the nested type for the new property as:

```typescript
interface RegistryEventNotificationConfig {
    pubsubTopicName: string;
    subfolderMatches?: string;
}
```

And [our assert](https://github.com/pulumi/pulumi-terraform/blob/d92dd348f40dff2873c69359ca21864a7b2f8148/pkg/tfgen/generate_nodejs.go#L1624-L1625) fails because we're trying to create another nested type with the same name but different declarations.

The reason the type names are the same is because we use a _singular_ name for properties that are arrays, which in this case happens to conflict with the nested type that already exists for the deprecated _singular_ property.

We'll need to do the following:
 1. Provide some mechanism to manually override the naming
 2. Come up with some convention to use when we can't use the singular name for arrays. Perhaps the manual overridden name we use for this will be to append an `Item` (so `RegistryEventNotificationConfigItem`)?

@lukehoban, do you have any preference on the naming scheme here?

stack72:
Fixes #212


resources.go
```diff
@@ -843,7 +855,21 @@ func Provider() tfbridge.ProviderInfo {
...
+				Tok: gcpResource(gcpKMS, "Registry"),
+				Fields: map[string]*tfbridge.SchemaInfo{
+					// This property's nested type name conflicts with the nested type of the existing (now deprecated)
+					// `event_notification_config` property (singular, a TypeMap). A conflict occurs because the new
+					// `event_notification_configs` property (plural, a TypeList) is a TypeList, which we singularize.
+					// To avoid the conflict, we override the nested type name for the new property, appending an "Item"
+					// suffix.
+					"event_notification_configs": {
+						Elem: &tfbridge.SchemaInfo{
+							NestedType: "RegistryEventNotificationConfigItem",
```
justinvp: @lukehoban, just to confirm, are you OK with using the `Item` suffix for this? I considered other names like `Entry` and `Element` but didn't feel too strongly about any of them. Open to changing it or other ideas.

Using a suffix like this seems in alignment with the place you [pointed out](https://github.com/pulumi/pulumi-gcp/issues/212#issuecomment-528487118) where we used a `Collection` suffix.

PR #224, opened 2019-09-16
Setting gcloud auth to use Python2.7 rather than the global use of 3.x

stack72:


stack72:
Works as expected - including in #222

Issue #229, opened 2019-10-04
typo: MangedSslCertificate


gorilych:
subj. should be Man**a**gedSslCertificate

```
pulumi-gcp$ grep -ir mangedsslcertificate | wc -l
60
pulumi-gcp$ grep -ir managedsslcertificate | wc -l
0
```

PR #230, opened 2019-10-04
Fix typo [Mm]angedSslCertificate -> [Mm]anagedSslCertificate

gorilych:
Fix for issue #229

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ CHANGELOG
 
 ## HEAD (Unreleased)
 * Regenerate SDK against tf2pulumi 0.6.0
+* Fix typo in resource name ManagedSslCertificate
 
 ---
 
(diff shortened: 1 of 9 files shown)


mikhailshilkov:
Hey @gorilych thanks for spotting this issue and opening a PR!
To avoid breaking changes, we prefer to keep the old resource with a deprecation notice and add a new one as an alias.

There's an example of this in the [Azure provider](https://github.com/pulumi/pulumi-azure/blob/master/resources.go#L1527-L1529). You'd need to "rename" the resource, generate SDKs, and make sure both resources are generated. I'm sure @stack72 can help here.

stack72:
@gorilych 

please accept my apologies - I didn't see this PR and just submitted #231 as a fix to your issue in #229 

Thank you so much for this

Paul

PR #231, opened 2019-10-04
Rename MangedSslCertificate to ManagedSslCertificate

stack72:
Fixes: #229

### DIFF
--- a/resources.go
+++ b/resources.go
@@ -493,7 +493,6 @@ func Provider() tfbridge.ProviderInfo {
 			},
 			"google_compute_instance_template":       {Tok: gcpResource(gcpCompute, "InstanceTemplate")},
 			"google_compute_interconnect_attachment": {Tok: gcpResource(gcpCompute, "InterconnectAttachment")},
-			"google_compute_managed_ssl_certificate": {Tok: gcpResource(gcpCompute, "MangedSslCertificate")},
 			"google_compute_node_group":              {Tok: gcpResource(gcpCompute, "NodeGroup")},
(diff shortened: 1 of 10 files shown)


resources.go
```diff
@@ -1501,3 +1502,84 @@ func Provider() tfbridge.ProviderInfo {
...
+	//	if legacyInfo.Docs == nil {
+	//		legacyInfo.Docs = &tfbridge.DocInfo{
+	//			Source: tfName[len("azurerm_"):] + ".html.markdown",
+	//		}
+	//	}
+	//
+	//	prov.DataSources[tfName] = &currentInfo
+	//	prov.DataSources[legacyTFName] = &legacyInfo
+	//	prov.P.DataSourcesMap[legacyTFName] = prov.P.DataSourcesMap[tfName]
+	//}
```
mikhailshilkov: Should we delete these lines?

stack72: I added these specifically because of that's the func to rename DataSources. I am going to investigate moving this into `pulumi-terraform-bridge` at some point. But as of now, it's not hugely understandable how we do this

stack72:
This passes on go1.12.9 - I am fine with that for now - we will be relooking at go1.13.x later

resources.go
mikhailshilkov: I referred to the commented out lines

stack72: Yeah - we have 2 funcs, 1 to rename resources (which is used) and 1 to rename DataSources, which is what you are referring to

This is only there for now until we can move this back to tfprovider - it will disappear soon:tm:

PR #233, opened 2019-10-09
Upgrade to v2.17.0 of the Google Beta Terraform Provider

stack72:


go.mod.old
```diff
@@ -0,0 +1,26 @@
+module github.com/pulumi/pulumi-gcp
```
jen20: We don't need to commit this file anymore

PR #234, opened 2019-10-11
Examples: Change the examples to run as individual tests

stack72:


examples/examples_test.go
```diff
@@ -28,120 +28,191 @@ import (
...
+
+	return cwd
+}
+
+func getBaseOptions(t *testing.T) integration.ProgramTestOptions {
+	project := getProject(t)
+	zone := getZone(t)
+	region := getRegion(t)
+	return integration.ProgramTestOptions{
+		ExpectRefreshChanges: true,
```
lukehoban: Similar to other PR - let’s not functionally change the tests here. Adding this non-trivially reduces the strictness of the tests - and we probably shouldn’t do that unless we need to (and for known/expected reason).

stack72: And this is why we have reviews :) Good pick up!

PR #240, opened 2019-10-28
Upgrade to v2.28.1 of the Google Beta Terraform Provider

stack72:


stack72:
Self merging so we can get this point release out - this was causing breakages for some people upstream

PR #257, opened 2019-12-02
Add a test for a C# example

mikhailshilkov:


mikhailshilkov:
This requires the 1.4.3 release of pulumi-terraform-bridge to build successfully

PR #263, opened 2019-12-12
Upgrade to v1.5.2 of pulumi-terraform-bridge

stack72:


stack72:
test failure unrelated - this is just a sporadic failure that we hit sometimes

PR #276, opened 2020-01-28
Merge pull request #276 from pulumi/stack72/bridge-v1.6.4

stack72:


.golangci.yml
```diff
@@ -4,9 +4,9 @@ linters:
     - deadcode
     - errcheck
     - goconst
-    - gofmt
     - gosec
     - govet
+    - gofmt
```
mikhailshilkov: Same question about the order

PR #307, opened 2020-03-26
Adapt package publishing for new repo layout

stack72:
This is required due to the new go mod module layout

Also update the version of the Google API as this repo was pulling in
a later version - this is what was breaking the PR


stack72:
//cc @EvanBoyle

scripts/go.mod
```diff
@@ -2,4 +2,4 @@ module github.com/pulumi/pulumi-gcp/scripts
 
 go 1.13
 
-require github.com/pulumi/pulumi/pkg v0.0.0-20200323215703-1d4a75674b2e
+require github.com/pulumi/pulumi/pkg v0.0.0-20200322194843-61928f04e052
```
mikhailshilkov: This seems to be going back? Is that intended? Why so many changes in this PR?

stack72: The original PR had picked up an incorrect version of the upstream API causing it to throw a build failure

The issue was resolved by using a replace in the go.mod

This version of pulumi/pkg makes everything the same version in the report now

sdk/go/gcp/accesscontextmanager/pulumiManifest.go
```diff
@@ -0,0 +1,13 @@
+// *** WARNING: this file was generated by the Pulumi Terraform Bridge (tfgen) Tool. ***
```
mikhailshilkov: I think you should gitignore these files

stack72: yeah I am just doing that right now :)

EvanBoyle:
Awesome, thanks for picking this up!

PR #312, opened 2020-03-30
Bump modules to v2 structure

stack72:
This is required so that users can run go get and the version without
needing to know the individual Git SHA

### DIFF
--- a/scripts/publish-plugin.sh
+++ b/scripts/publish-plugin.sh
@@ -17,7 +17,7 @@ if [ "$(go env GOOS)" = "windows" ]; then
 fi
 
 (cd "${ROOT}/provider" && go build \
-   -ldflags "-X github.com/pulumi/pulumi-gcp/provider/pkg/version.Version=${VERSION}" \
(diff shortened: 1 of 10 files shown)


scripts/publish-plugin.sh
```diff
@@ -17,7 +17,7 @@ if [ "$(go env GOOS)" = "windows" ]; then
 fi
 
 (cd "${ROOT}/provider" && go build \
-   -ldflags "-X github.com/pulumi/pulumi-gcp/provider/pkg/version.Version=${VERSION}" \
+   -ldflags "-X github.com/pulumi/pulumi-gcp/provider/v2/pkg/version.Version=${VERSION}" \
    -o "${WORK_PATH}/pulumi-resource-gcp${BIN_SUFFIX}" \
    "${ROOT}/cmd/pulumi-resource-gcp")
```
EvanBoyle: Have you tested this code path? I wouldn't be surprised if that last line needs to be updated. It looks like that's a package?

stack72: I was able to run this exact script from my local machine and it works

PR #317, opened 2020-04-06
Re-Enable linting of provider as part of CI

stack72:
Related: https://github.com/pulumi/pulumi/issues/3834


stack72:
Handling as part of move to GHA

PR #324, opened 2020-04-20
Upgrade to v3.17.0 of the Google Beta Terraform Provider

stack72:


sdk/dotnet/Kms/Registry.cs
```diff
@@ -9,63 +9,32 @@
 
 namespace Pulumi.Gcp.Kms
 {
-    /// <summary>
-    ///  Creates a device registry in Google's Cloud IoT Core platform. For more information see
-    /// [the official documentation](https://cloud.google.com/iot/docs/) and
-    /// [API](https://cloud.google.com/iot/docs/reference/cloudiot/rest/v1/projects.locations.registries).
-    /// </summary>
     public partial class Registry : Pulumi.CustomResource
```
mikhailshilkov, line 12: Don't we mark renamed resources as `[Obsolete]`?

stack72: We do not as of this moment - it's something you added an issue for in the bridge https://github.com/pulumi/pulumi-terraform-bridge/issues/133

mikhailshilkov: Sorry, I keep forgetting. We should prioritize that task.

PR #326, opened 2020-04-20
BucketCor -> BucketCors

### LINKED ISSUE #325: Pulumi.Gcp.Storage.Inputs.BucketCorsArgs is now BucketCorArgs
In the .net API, BucketCorsArgs has had an 's' removed that wasn't from an unnecessary plural.

mikhailshilkov:
Resolves #325 (for all languages)
Let's discuss if we are comfortable doing that and how, as it's a breaking change.


stack72:
Going to close this out as we are tracking the bug in the bridge that we should fix rather than making it specific to this repo

PR #377, opened 2020-06-27
Fix minor type issue in the example code for sdk/go/gcp/compute/instance.go

albert-zhong:
I was playing around with the NewInstance function in [sdk/go/gcp/compute/instance.go](https://github.com/pulumi/pulumi-gcp/blob/master/sdk/go/gcp/compute/instance.go) and I believe I ran into 3 minor issues with the example code:

1.  Should be `:=` instead of `=` on line 30
2. The `Metadata` parameter should be of type `pulumi.StringMap` instead of `pulumi.Map` on line 37
3. `InstanceNetworkInterfaceArgs` should take in an `AccessConfigs` field (with a s at the end), with a type of `InstanceNetworkInterfaceAccessConfigArray`, which itself takes in `InstanceNetworkInterfaceAccessConfigArgs`

### DIFF
--- a/sdk/go/gcp/compute/instance.go
+++ b/sdk/go/gcp/compute/instance.go
@@ -18,56 +18,49 @@ import (
 // ## Example Usage
 //
 // ```go
-// package main
-//
-// import (
-// 	"github.com/pulumi/pulumi-gcp/sdk/v3/go/gcp/compute"
-// 	"github.com/pulumi/pulumi/sdk/v2/go/pulumi"
-// )
-//
 // func main() {
-// 	pulumi.Run(func(ctx *pulumi.Context) error {
-// 		_, err = compute.NewInstance(ctx, "default", &compute.InstanceArgs{
-// 			BootDisk: &compute.InstanceBootDiskArgs{
-// 				InitializeParams: &compute.InstanceBootDiskInitializeParamsArgs{
-// 					Image: pulumi.String("debian-cloud/debian-9"),
-// 				},
-// 			},
-// 			MachineType: pulumi.String("n1-standard-1"),
-// 			Metadata: pulumi.Map{
-// 				"foo": pulumi.String("bar"),
-// 			},
-// 			MetadataStartupScript: pulumi.String("echo hi > /test.txt"),
-// 			NetworkInterfaces: compute.InstanceNetworkInterfaceArray{
-// 				&compute.InstanceNetworkInterfaceArgs{
-// 					AccessConfig: pulumi.MapArray{
-// 						nil,
-// 					},
-// 					Network: pulumi.String("default"),
-// 				},
-// 			},
-// 			ScratchDisks: compute.InstanceScratchDiskArray{
-// 				&compute.InstanceScratchDiskArgs{
-// 					Interface: pulumi.String("SCSI"),
-// 				},
-// 			},
-// 			ServiceAccount: &compute.InstanceServiceAccountArgs{
-// 				Scopes: pulumi.StringArray{
-// 					pulumi.String("userinfo-email"),
-// 					pulumi.String("compute-ro"),
-// 					pulumi.String("storage-ro"),
-// 				},
-// 			},
-// 			Tags: pulumi.StringArray{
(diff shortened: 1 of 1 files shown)


lukehoban:
These examples are auto-generated - so we will need to fix the generator and regenerate them. 

Quite a lot of improvements have been made to the Go code generation recently - primarily tracked in https://github.com/pulumi/pulumi/issues/4330. 

@evanboyle Are the specific issues here things that are already fixed or that are tracked elsewhere?

albert-zhong:
> These examples are auto-generated - so we will need to fix the generator and regenerate them.
> 
> Quite a lot of improvements have been made to the Go code generation recently - primarily tracked in [pulumi/pulumi#4330](https://github.com/pulumi/pulumi/issues/4330).
> 
> @EvanBoyle Are the specific issues here things that are already fixed or that are tracked elsewhere?

I understand now, thanks! Also in the future I'll open a new issue for these kind of small bugs instead of a PR.

EvanBoyle:
Thanks for digging into this @albert-zhong!

1. The incorrect usage of `:=` is a known issue https://github.com/pulumi/pulumi/issues/4900
2. Improved typing for Maps is currently out for review in this PR if you're curious: https://github.com/pulumi/pulumi/pull/4914
3. Unfortunately our generated examples are only as good as the upstream source. Which at the moment has a bug (like mention it should be `AccessConfigs`. The typo also breaks the type inference for nested properties which is why we see `pulumi.Map` rather than the expected types that you listed. https://github.com/terraform-providers/terraform-provider-google/blame/master/website/docs/r/compute_instance.html.markdown#L42

For that last case, our best course of action is to open a PR in the upstream repo with a fix.

lukehoban:
> like mention it should be `AccessConfigs`

I believe the example is right, but that we are not picking up the appropriate transformation to an array in the Go codegen? (The tfbridge layer introduces the pluralization on the name - but somehow the Go codegen identity even picking up that it’s an array?)

EvanBoyle:
@lukehoban the go codegen checks against the schema which expects an `accessConfigs` property:

```json
        "gcp:compute/InstanceNetworkInterface:InstanceNetworkInterface": {
            "properties": {
                "accessConfigs": {
                    "type": "array",
                    "items": {
                        "$ref": "#/types/gcp:compute/InstanceNetworkInterfaceAccessConfig:InstanceNetworkInterfaceAccessConfig"
                    },
                    "description": "Access configurations, i.e. IPs via which this\ninstance can be accessed via the Internet. Omit to ensure that the instance\nis not accessible from the Internet. If omitted, ssh will not\nwork unless this provider can send traffic to the instance's network (e.g. via\ntunnel or because it is running on another cloud instance on that network).\nThis block can be repeated multiple times. Structure documented below.\n",
                    "language": {
                        "python": {
                            "mapCase": false
                        }
                    }
                },
```

EvanBoyle:
@albert-zhong I pushed a fix to another inflight PR for the assignment issue: https://github.com/pulumi/pulumi/pull/4915/commits/771bfb58de24912cd4615d9879dba160e4aa0346

You can see the assignment is fixed in the diff:
https://github.com/pulumi-bot/pulumi-gcp/blob/57011d872934135ce4bbad3a967b73d48b9ca61d/sdk/go/gcp/compute/instance.go#L30

lukehoban:
> which expects an accessConfigs property

The Pulumi name is `accessConfigs` but the upstream name is `access_config`.  So the upstream example is correct. 

Note that the TypeScript example generated does have the right name, but C# codegen appears to also be wrong in a similar way to Go?

EvanBoyle:
Python is missing it as well: https://github.com/pulumi/pulumi-gcp/blob/master/sdk/python/pulumi_gcp/compute/instance.py#L326

Not obvious to me where this is handled in the node generator.

EvanBoyle:
After talking to Pat, it looks like this is being done in tf2pulumi using some logic in the bridge. The nodejs examples for GCP look like they're still generated by the old codgen. Issue tracking moving this logic in to the pulumi/pulumi generators:

https://github.com/pulumi/pulumi/issues/4926

PR #388, opened 2020-07-14
Upgrade to v3.30.0 of the Google Beta Terraform Provider

stack72:
Fixes: #387


stack72:
> So many new examples!

I wish! That was due to the tfgen binary being on my path :/ It's back to normal now

PR #441, opened 2020-09-08
Upgrade to v3.38.0 of the Google Beta Terraform Provider

stack72:
Fixes: #440


sdk/dotnet/Notebooks/Instance.cs
```diff
@@ -336,15 +336,21 @@ public sealed class InstanceArgs : Pulumi.ResourceArgs
...
-        /// The owner of this instance after creation.
+        /// The list of owners of this instance after creation.
         /// Format: alias@example.com.
         /// Currently supports one owner only.
         /// If not specified, all of the service account users of
         /// your VM instance's service account can use the instance.
         /// </summary>
-        [Input("instanceOwners")]
-        public Input<string>? InstanceOwners { get; set; }
+        public InputList<string> InstanceOwners
```
mikhailshilkov, line 349: That's a breaking change here

mikhailshilkov (approved):
There is one breaking change in there. Let's apply our triage policy (and probably release it with a note in the notes?)

sdk/dotnet/Notebooks/Instance.cs
stack72: Oh, we can mitigate this with MaxItems: 1, the comment even says it only supports one owner

I will fix this up now

PR #453, opened 2020-09-29
Upgrade to v3.41.0 of the Google Beta Terraform Provider

stack72:
Fixes: #452


sdk/dotnet/CloudBuild/Inputs/TriggerBuildArtifactsObjectsArgs.cs
```diff
@@ -32,13 +32,19 @@ public InputList<string> Paths
...
+        private InputList<Inputs.TriggerBuildArtifactsObjectsTimingArgs>? _timings;
+
         /// <summary>
         /// -
         /// Output only. Stores timing information for pushing all artifact objects.
         /// Structure is documented below.
         /// </summary>
-        [Input("timing")]
-        public Input<Inputs.TriggerBuildArtifactsObjectsTimingArgs>? Timing { get; set; }
+        public InputList<Inputs.TriggerBuildArtifactsObjectsTimingArgs> Timings
```
justinvp, line 43: This (along with many other similar changes in this PR) is a breaking change. Are we bumping the major version before release? Or do we need to revert this back to how it was previously?

stack72: There was no plans to bump the version on this - this is actually what changed upstream in this minor release. We might be able to use MaxItems to remediate if the API doesn't actually take more than 1 item in this case

There are other times we've accepted a breaking change to these providers when things like this happen

//cc @leezen

mikhailshilkov: Yeah, agrees with Justin here, the changes are quite significant. Are they all `MaxItems=1`-able?

sdk/nodejs/cloudrun/domainMapping.ts
```diff
@@ -71,7 +71,7 @@ export class DomainMapping extends pulumi.CustomResource {
     /**
      * The current status of the DomainMapping.
      */
-    public /*out*/ readonly status!: pulumi.Output<outputs.cloudrun.DomainMappingStatus>;
+    public /*out*/ readonly statuses!: pulumi.Output<outputs.cloudrun.DomainMappingStatus[]>;
```
stack72, line 74: This is an output property that has no MaxItems set on it - breaking

sdk/nodejs/cloudrun/service.ts
```diff
@@ -87,7 +87,7 @@ export class Service extends pulumi.CustomResource {
     /**
      * The current status of the Service.
      */
-    public /*out*/ readonly status!: pulumi.Output<outputs.cloudrun.ServiceStatus>;
+    public /*out*/ readonly statuses!: pulumi.Output<outputs.cloudrun.ServiceStatus[]>;
```
stack72, line 90: This is a computed property upstream that has no MaxItems set on it - breaking

sdk/nodejs/compute/interconnectAttachment.ts
```diff
@@ -123,7 +123,7 @@ export class InterconnectAttachment extends pulumi.CustomResource {
      * Information specific to an InterconnectAttachment. This property is populated if the interconnect that this is attached
      * to is of type DEDICATED.
      */
-    public /*out*/ readonly privateInterconnectInfo!: pulumi.Output<outputs.compute.InterconnectAttachmentPrivateInterconnectInfo>;
+    public /*out*/ readonly privateInterconnectInfos!: pulumi.Output<outputs.compute.InterconnectAttachmentPrivateInterconnectInfo[]>;
```
stack72, line 126: This is a computed property upstream that has no MaxItems set on it - breaking

sdk/nodejs/container/getCluster.ts
```diff
@@ -45,29 +43,19 @@ export interface GetClusterArgs {
      * is not provided, the provider project is used.
      */
     readonly project?: string;
-    /**
-     * The region this cluster has been created in. Deprecated
-     * in favour of `location`.
-     */
-    readonly region?: string;
```
stack72, line 52: Removed in the upstream provider as it was previously deprecated

sdk/nodejs/datacatalog/entry.ts
```diff
@@ -54,11 +54,11 @@ export class Entry extends pulumi.CustomResource {
      * Specification for a group of BigQuery tables with name pattern [prefix]YYYYMMDD. Context:
      * https://cloud.google.com/bigquery/docs/partitioned-tables#partitioning_versus_sharding.
      */
-    public /*out*/ readonly bigqueryDateShardedSpec!: pulumi.Output<outputs.datacatalog.EntryBigqueryDateShardedSpec>;
+    public /*out*/ readonly bigqueryDateShardedSpecs!: pulumi.Output<outputs.datacatalog.EntryBigqueryDateShardedSpec[]>;
```
stack72, line 57: This is a computed property upstream that has no MaxItems set on it - breaking

sdk/nodejs/datacatalog/entry.ts
```diff
@@ -54,11 +54,11 @@ export class Entry extends pulumi.CustomResource {
...
     /**
      * Specification that applies to a BigQuery table. This is only valid on entries of type TABLE.
      */
-    public /*out*/ readonly bigqueryTableSpec!: pulumi.Output<outputs.datacatalog.EntryBigqueryTableSpec>;
+    public /*out*/ readonly bigqueryTableSpecs!: pulumi.Output<outputs.datacatalog.EntryBigqueryTableSpec[]>;
```
stack72, line 61: This is a computed property upstream that has no MaxItems set on it - breaking

sdk/nodejs/dataproc/job.ts
```diff
@@ -107,7 +107,7 @@ export class Job extends pulumi.CustomResource {
     /**
      * The status of the job.
      */
-    public /*out*/ readonly status!: pulumi.Output<outputs.dataproc.JobStatus>;
+    public /*out*/ readonly statuses!: pulumi.Output<outputs.dataproc.JobStatus[]>;
```
stack72, line 110: This is a computed property upstream that has no MaxItems set on it - breaking

sdk/nodejs/iot/device.ts
```diff
@@ -52,7 +52,7 @@ export class Device extends pulumi.CustomResource {
     /**
      * The most recent device configuration, which is eventually sent from Cloud IoT Core to the device.
      */
-    public /*out*/ readonly config!: pulumi.Output<outputs.iot.DeviceConfig>;
+    public /*out*/ readonly configs!: pulumi.Output<outputs.iot.DeviceConfig[]>;
```
stack72, line 55: This is a computed property upstream that has no MaxItems set on it - breaking

sdk/nodejs/iot/device.ts
```diff
@@ -116,7 +116,7 @@ export class Device extends pulumi.CustomResource {
     /**
      * The state most recently received from the device.
      */
-    public /*out*/ readonly state!: pulumi.Output<outputs.iot.DeviceState>;
+    public /*out*/ readonly states!: pulumi.Output<outputs.iot.DeviceState[]>;
```
stack72, line 119: This is a computed property upstream that has no MaxItems set on it - breaking

sdk/nodejs/iot/device.ts
```diff
@@ -74,7 +74,7 @@ export class Device extends pulumi.CustomResource {
     /**
      * The error message of the most recent error, such as a failure to publish to Cloud Pub/Sub.
      */
-    public /*out*/ readonly lastErrorStatus!: pulumi.Output<outputs.iot.DeviceLastErrorStatus>;
+    public /*out*/ readonly lastErrorStatuses!: pulumi.Output<outputs.iot.DeviceLastErrorStatus[]>;
```
stack72, line 77: This is a computed property upstream that has no MaxItems set on it - breaking

stack72:
@leezen / @justinvp / @mikhailshilkov so I have been through all the fields that are *breaking* and they are actual breakages. These are things that changes in the upstream API specs that the module gets generated from (i.e. MagicModules)

I think there are going to be more of these in the coming weeks as they are all being attributes to "refactoring"

I am not sure what is the best way to proceed with this and would like some input - this *should* be a breaking change and thus a major version bump...

leezen:
@stack72 A couple thoughts:
1) Give the number and impact of the breaking changes, I do think this will need to be a major version bump.
2) Are there any significant functionality improvements in this release? If not, would it make sense to hold a week or two longer to capture more of the breaking changes you expect to come?

PR #469, opened 2020-10-29
Upgrade to v3.45.0 of the Google Beta Terraform Provider

stack72:
Fixes: #467


provider/resources.go
```diff
@@ -214,6 +214,9 @@ func Provider() tfbridge.ProviderInfo {
 					},
 				},
 			},
+			"google_access_context_manager_access_level_condition": {
+				Tok: gcpResource(gcpAccessContextManager, "AccessLevelsCondition"),
```
mikhailshilkov, line 218: Should be "level"?
```suggestion
				Tok: gcpResource(gcpAccessContextManager, "AccessLevelCondition"),
```

stack72: Fixed and a rebuild is underway! Great catch

mikhailshilkov (approved):

PR #480, opened 2020-11-17
Upgrade to v3.48.0 of the Google Beta Terraform Provider

stack72:
Fixes: #478


sdk/dotnet/Iam/GetWorkloadIdentityPool.cs
```diff
@@ -18,15 +18,6 @@ public static Task<GetWorkloadIdentityPoolResult> InvokeAsync(GetWorkloadIdentit
...
     public sealed class GetWorkloadIdentityPoolArgs : Pulumi.InvokeArgs
     {
-        [Input("description")]
-        public string? Description { get; set; }
-
-        [Input("disabled")]
-        public bool? Disabled { get; set; }
-
-        [Input("displayName")]
-        public string? DisplayName { get; set; }
```
mikhailshilkov, line 28: That's a breaking change code-wise - but it looks like a bug fix?

mikhailshilkov (approved):
LGTM except for one fishy change - pls confirm it's intended

sdk/dotnet/Iam/GetWorkloadIdentityPool.cs
stack72: So these values were never set and were only introduced last week in the upgrade to v3.47.0

I am going to apply the rule of thumb that this will affect < 1% of users so we are ok for this to be merged

PR #489, opened 2020-12-09
Upgrade to v3.50.0 of the Google Beta Terraform Provider

stack72:
Fixes: #488
Fixes: #495


sdk/dotnet/Compute/GetResourcePolicy.cs
```diff
@@ -66,42 +66,43 @@ public GetResourcePolicyArgs()
     [OutputType]
     public sealed class GetResourcePolicyResult
     {
-        /// <summary>
-        /// Description of this Resource Policy.
-        /// </summary>
-        public readonly string Description;
```
mikhailshilkov, line 72: `Description` property is gone?

stack72: This is actually a property that never existed upstream in the SDK and thus was never actually set so it's more of a cleanup on the TF side

mikhailshilkov (approved):

PR #531, opened 2021-02-24
Upgrade to v3.58.0 of the Google Beta Terraform Provider

stack72:
Fixes: #530


sdk/dotnet/Compute/TargetSSLProxy.cs
```diff
@@ -146,11 +146,11 @@ public partial class TargetSSLProxy : Pulumi.CustomResource
...
         /// <summary>
         /// A list of SslCertificate resources that are used to authenticate
-        /// connections between users and the load balancer. Currently, exactly
-        /// one SSL certificate must be specified.
+        /// connections between users and the load balancer. At least one
+        /// SSL certificate must be specified.
         /// </summary>
         [Output("sslCertificates")]
-        public Output<string> SslCertificates { get; private set; } = null!;
+        public Output<ImmutableArray<string>> SslCertificates { get; private set; } = null!;
```
mikhailshilkov, line 153: Looks like a `MaxItems` breaking change

stack72: @mikhailshilkov:

"compute: removed max items limit from google_compute_target_ssl_proxy. The API currently allows upto 15 Certificates." from their changelog...

mikhailshilkov: Do we want to revert this until the next major version? Or does it pass our "not too impacting but fixes correctness" filter?

stack72: I, personally, believe this would fall into our "affects less than 5% of users" category - but always willing to move from that stance

mikhailshilkov (approved):

PR #537, opened 2021-03-09
Upgrae to v3.59.0 of the Google Beta Terraform Provider

stack72:
Fixes: #536


provider/resources.go
```diff
@@ -918,6 +920,7 @@ func Provider() tfbridge.ProviderInfo {
 				},
 			},
 			"google_dataproc_autoscaling_policy": {Tok: gcpResource(gcpDataProc, "AutoscalingPolicy")},
+			"google_dataproc_metastore_service":  {Tok: gcpResource(gcpDataProc, "MetaStoreService")},
```
mikhailshilkov, line 923: It seems to be "Metastore" in the docs https://cloud.google.com/dataproc-metastore/docs

sdk/dotnet/Redis/Instance.cs
```diff
@@ -330,12 +330,6 @@ public sealed class InstanceArgs : Pulumi.ResourceArgs
         [Input("authEnabled")]
         public Input<bool>? AuthEnabled { get; set; }
 
-        /// <summary>
-        /// AUTH String set on the instance. This field will only be populated if auth_enabled is true.
-        /// </summary>
-        [Input("authString")]
-        public Input<string>? AuthString { get; set; }
```
mikhailshilkov, line 337: Breaking change?

stack72: This is expected - this was never a valid input - it's a computed field only

provider/resources.go
stack72: Fixed!

mikhailshilkov (approved):

PR #551, opened 2021-03-30
Upgrade to v3.62.0 of the Google Beta Terraform Provider

stack72:
Fixes: #550


sdk/dotnet/Compute/InterconnectAttachment.cs
```diff
@@ -490,7 +490,7 @@ public InputList<string> CandidateSubnets
         /// this interconnect attachment. Currently, only 1440 and 1500 are allowed. If not specified, the value will default to 1440.
         /// </summary>
         [Input("mtu")]
-        public Input<string>? Mtu { get; set; }
+        public Input<int>? Mtu { get; set; }
```
mikhailshilkov, line 493: These changes are breaking. Should we revert in the fork? Or does it fix a bug?

stack72: From their CHANGELOG:

> compute: corrected underlying type to integer for field mtu in google_compute_interconnect_attachment

mikhailshilkov: Was it working at all before? If so, is that a non-breaking change for them but a breaking change for us?

stack72: From what I can see it's a pure bug fix - it wasn't handled correctly in their code but now it is

mikhailshilkov: Okay then

mikhailshilkov (approved):

PR #566, opened 2021-04-21
Upgrade to v3.65.0 of the Google Beta Terraform Provider

stack72:
Fixes: #565

Please Note: gcp:eventarc.Trigger has had a change that adds support
for specifying transports rather than them only being a computed property


mikhailshilkov:
> Please Note: gcp:eventarc.Trigger has had a change that adds support
> for specifying transports rather than them only being a computed property

I'm not sure I understand this exactly, but the change looks bad enough (broken input and output properties). Any way to avoid this in our fork? Overall, I think it's worth triaging.

stack72:
ok @mikhailshilkov latest changes look better! I have noted the changes we introduced to be reverted for the [next breaking change](https://github.com/pulumi/pulumi-gcp/issues/567):

```
▶ schema-tools compare gcp master --local-path=provider/cmd/pulumi-resource-gcp/schema.json
Looking good! No breaking changes found.

New resources/functions:
gcp:kms/getKMSSecretAsymmetric:getKMSSecretAsymmetric
```

leezen (approved):

PR #583, opened 2021-05-27
Upgrade to v3.2.1 of the pulumi-terraform-bridge

stack72:


stack72:
Self merging as this is part of a series that was centrally reviewed

PR #617, opened 2021-07-13
Upgrade to v3.75.0 of the Google Beta Terraform Provider

stack72:
Fixes: #616﻿


provider/resources.go
```diff
@@ -1932,22 +1932,23 @@ func Provider() tfbridge.ProviderInfo {
 			// Certificate Authority
 			"google_privateca_certificate_authority": {Tok: gcpResource(gcpCertificateAuthority, "Authority")},
 			"google_privateca_certificate":           {Tok: gcpResource(gcpCertificateAuthority, "Certificate")},
-			"google_privateca_certificate_authority_iam_binding": {
-				Tok: gcpResource(gcpCertificateAuthority, "AuthorityIamBinding"),
+			"google_privateca_ca_pool":               {Tok: gcpResource(gcpCertificateAuthority, "CaPool")},
```
mikhailshilkov, line 1935: Do we already use `Ca` capitalization elsewhere? If not, it may be more natural to use `CA` instead (`CAPool`).

mikhailshilkov (approved):
LGTM with one suggestion

provider/resources.go
stack72: We've been using Ca in other places as well - thank you for checking it!

PR #657, opened 2021-10-01
Update bridge

komalali:


sdk/go/gcp/bigquery/datasetAccess.go
```diff
@@ -14,7 +14,7 @@ import (
 // ## Import
 //
 // This resource does not support import.
-type DatasetAccessType struct {
+type DatasetAccess struct {
```
justinvp, line 17: Are these changes expected?

komalali: These are related to https://github.com/pulumi/pulumi/pull/7985 and you can see them in the [downstream tests](https://github.com/pulumi-bot/pulumi-gcp/compare/f8f3c9008..389a523c3#diff-97f99b26a6fb54d526e6d8d281f8fffffd136350adcbd14148c985b86fa3dd52R17) there.

komalali: And yeah it looks correct now and was wrong before. Since `DataAccess` is a resource and it was being incorrectly named `DataAccessType`. I think this might've been a regression introduced by https://github.com/pulumi/pulumi/pull/7857

komalali: Indeed, it was a regression that is now corrected. https://github.com/pulumi-bot/pulumi-gcp/compare/bcc0bc857..e3dfca49d#diff-97f99b26a6fb54d526e6d8d281f8fffffd136350adcbd14148c985b86fa3dd52R17

pgavlin (approved):

PR #679, opened 2021-11-01
[DOC] fix readme install with pip

raphaelauv:
fix with correct package name https://pypi.org/project/pulumi-gcp/

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@ or `yarn`:
 
 To use from Python, install using `pip`:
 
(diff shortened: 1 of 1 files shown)


leezen:
Both forms `pulumi-gcp` and `pulumi_gcp` should work, though some reading suggests the hyphen is more canonical. I think if we want to do this, we should likely do this for all packages and not just `pulumi-gcp` so that it's consistent. @stack72 Any particular opinions?

PR #708, opened 2021-12-08
Fix: code examples for gcp-compute-subnetwork

la3mmchen:
Hi,

I hope the code i touched is the part that your docs that is displayed here: https://www.pulumi.com/registry/packages/gcp/api-docs/compute/subnetwork/#example-usage

Without those modifications the displayed go code won't work as there is no object "custom_test".

Bye

### DIFF
--- a/sdk/go/gcp/compute/subnetwork.go
+++ b/sdk/go/gcp/compute/subnetwork.go
@@ -54,7 +54,7 @@ import (
 //
 // func main() {
 // 	pulumi.Run(func(ctx *pulumi.Context) error {
-// 		_, err := compute.NewNetwork(ctx, "custom_test", &compute.NetworkArgs{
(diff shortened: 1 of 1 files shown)


stack72 (commented):
Hi @la3mmchen 

Thanks so much for taking the time to update these - unfortunately these code comment examples are autogenerated and anything we manually change in these files will be overwritten by our next generation

The reason the examples use _ is that we don't use the resource name and therefore we would not be generating valid code

I'm going to close this out so that there is no confusion as to why it continues to stay open

Sorry about this and thank you again for the effort you have put in

Paul

la3mmchen:
Nevermind @stack72 thanks for pointing it out.

Is your documentation somewhere in github? there is wrong code at https://www.pulumi.com/registry/packages/gcp/api-docs/compute/subnetwork/ is wrong and if you point me to the place i try to fix it again ;-)

PR #775, opened 2022-03-23
Fix documentation issue for serviceaccount

### LINKED ISSUE #580: Python import of `service_account` fails / documenation wrong
I am trying to create a Service Account on GCP with the pulumi python module according to the documentation. This fails with an import error

Following the example usage described in the documentation https://www.pulumi.com/docs/reference/pkg/gcp/serviceaccount/account/ 

```
import pulumi
import pulumi_gcp as gcp

service_account = gcp.service_account.Account("serviceAccount",
    account_id="service-account-id",
    display_name="Service Account")
```
results in this error:

```
  File "./__main__.py", line 24, in <module>
    service_account = gcp.service_account.Account(
AttributeError: module 'pulumi_gcp' has no attribute 'service_account'
```

I would expect this to properly create a service account

Versions: `pulumi-gcp==5.3.0`, pulumi version v3.2.1

Looking at the code https://github.com/pulumi/pulumi-gcp/blob/master/sdk/python/pulumi_gcp/__init__.py#L75 shows that underscores seem to be removed.

So the above code works when removing the underscore in the module as well. I am assuming this will fail for other modules as well.

I would propose to either fix the documentation or change the code to not remove the underscores

jdoss:
This fixes #580 by removing two underscores.

### DIFF
--- a/sdk/python/pulumi_gcp/serviceaccount/account.py
+++ b/sdk/python/pulumi_gcp/serviceaccount/account.py
@@ -297,7 +297,7 @@ class Account(pulumi.CustomResource):
         import pulumi
         import pulumi_gcp as gcp
 
-        service_account = gcp.service_account.Account("serviceAccount",
+        service_account = gcp.serviceaccount.Account("serviceAccount",
             account_id="service-account-id",
             display_name="Service Account")
         ```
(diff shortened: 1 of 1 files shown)


guineveresaenger:
Hi @jdoss - thank you for your pull request, and welcome!

Unfortunately we cannot accept this change as all files under `sdk/*` are generated by the Pulumi codegen library (including documentation), and dependent on an upstream provider, so these changes would revert on the next release of this provider.  Please vote on #580 with a 👍 on the issue to help us prioritize a fix.

Thank you again!

noahdatasherpas:
Hello, changing the code may not be acceptable but would it be possible to at least update the documentation?

susanev:
@noahdatasherpas i believe the 580 issue that Guin linked to is the tracking issue for the docs issue.

noahdatasherpas:
@susanev Yes. I commented on that issue as well.

PR #788, opened 2022-04-11
updating some links and link text in the readme

susanev:


susanev:
i have no idea why i opened two of these.

PR #797, opened 2022-04-22
fix: url mask for RegionNetworkEndpointGroupServerlessDeployment should be optional

knowhoper:


knowhoper:
As per: https://github.com/hashicorp/terraform-provider-google/issues/11242

Issue #869, opened 2022-08-09
Fix Debian image version for acceptance tests


guineveresaenger:
Cron runs are failing with `Error resolving image name 'debian-cloud/debian-9': Could not find image or family debian-cloud/debian-9` 

To fix, replace image names with `debian-11` to ensure viability of integration tests.

guineveresaenger:
Closing via #870

PR #870, opened 2022-08-09
Update Debian image for integration tests

guineveresaenger:
Fix for #869

### DIFF
--- a/examples/loadbalancer/index.ts
+++ b/examples/loadbalancer/index.ts
@@ -18,7 +18,7 @@ const regionZone = gcp.config.zone;
 const projectName = gcp.config.project;
 
 const machineTypeName = "f1-micro";
-const imageName = "debian-cloud/debian-9";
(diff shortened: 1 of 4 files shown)


sdk/dotnet/CertificateAuthority/CaPool.cs
```diff
@@ -43,6 +43,132 @@ namespace Pulumi.Gcp.CertificateAuthority
     /// 
     /// }
     /// ```
+    /// ### Privateca Quickstart
```
viveklak, line 46: Sounds like you only updated the docker image in the tests? Are these SDK changes expected?

guineveresaenger: I did regenerate the SDKs because I was debugging the tests so this is a bit odd - let me remove these

viveklak (approved):

PR #891, opened 2022-09-01
Use provider host client Log function to show warning to Pulumi output

### LINKED ISSUE #890: unable to find required configuration setting: GCP Project
The [new auth config validation](https://github.com/pulumi/pulumi-gcp/pull/882) is causing deployments to break that do not set the default `gcp:project` key.

Initial thoughts:

This chunk of code is too zealous and we can begin by removing it and roll the error into the gcloud config validation below:
https://github.com/pulumi/pulumi-gcp/blob/master/provider/resources.go#L187
That will fix things for users who have non-default settings.

We can then look into how the underlying provider works - exactly how the GCP project is required and under which circumstances

If necessary/unaccounted for in the auth helper library, we should add validation tests that take into account all authentication scenarios.

guineveresaenger:
Fixes #890

BLOCKED by https://github.com/pulumi/pulumi-terraform-bridge/pull/605 - must release a new version of the bridge and rebuild the provider

Sample output during `pulumi preview` will look something like this:

Screen Shot 2022-09-02 at 2 00 12 PM


guineveresaenger:
This is pulling in the changes from https://github.com/pulumi/pulumi-terraform-bridge/pull/605 and I've verified it works with a local build of this provider, as advertised above.

stack72 (approved):
Fantastic addition and seeing it through the bridge - thank you!!

PR #995, opened 2023-02-16
Surface access key validation error

rquitales:
Helps surface the issue behind #989 

This PR surfaces the error when validating access credentials fails. Previously, we assumed that validation fails because the GCP credentials are not provided and instructs the user to specify credentials. However, validation could fail for different reasons so this should also be surfaced to the user to provide more helpful error messages for debugging.


squaremo:
Oddly, the "Update changelog" check passed without an update and without the magic label. I think it's worth adding a changelog entry for this one.

kpitzen (approved):
Love this!

PR #1047, opened 2023-03-29
Allow specifying environment variables from secrets in generated CloudFunction

tjsousa:
There's a mechanism already in place to specify custom args to the auto-generated functions created as the result of callbacks defined for bucket's onObjectEvents, but unfortunately this option supported by CloudFunction is not whitelisted in the type.

This should be all that it takes for our use-case to be able to customize this value.

### DIFF
--- a/sdk/nodejs/cloudfunctions/zMixins.ts
+++ b/sdk/nodejs/cloudfunctions/zMixins.ts
@@ -354,6 +354,10 @@ export interface CallbackFunctionArgs {
      */
     region?: pulumi.Input<string>;
     /**
+     * Secret environment variables configuration. Structure is documented below.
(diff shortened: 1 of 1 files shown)


tjsousa:
Hey @kpitzen, I believe I was missing an import statement. Would you mind running tests again?

kpitzen:
Hi @tjsousa - apologies for the delay here! I'll get those running now.

kpitzen:
Hey @tjsousa - this all looks good on our end! I'll go ahead and get this merged in. Thank you for your contribution!

tjsousa:
Can confirm that this is working for us in GCP after integration of the latest release.

PR #1048, opened 2023-04-04
Upgrade v4.60.0

### LINKED ISSUE #1046: Upgrade terraform-provider-google-beta to v4.59.0
Release details: https://github.com/hashicorp/terraform-provider-google-beta/releases/tag/v4.59.0

danielrbradley:
Fixes #1046 

The v4.60.0 tag is not yet pushed but the release branch is already up - so should be able to jump straght to there.


danielrbradley:
The upstream diff shows that `terraform-plugin-framework` has started to be addopted which explains the missing data sources - these are the first to be migrated across. cc: @iwahbe

t0yv0:
What's your sense @danielrbradley about the missing functions. Should we prioritize getting these functions muxed in before doing a next release, or should we do a release anyway accepting that they are temporarily unavailable. Is there any usage data we should be consulting to influence this tradeoff?

danielrbradley:
@t0yv0 these are pretty important functions so can't leave them out. I've prepped a patch to revert the conversions in our fork for now for this release. Then I think we should prioritise integrating the muxer for the subsequent release.

iwahbe (approved):
Looks good!

Can you explain the logic behind mapping all the `google_dataplex_lake_iam_*` to `"dataplex_lake_iam.html.markdown"`? How should we know to do that?

danielrbradley:
@iwahbe I'm not too familiar with it either. 

tf_gen passed fine locally but failed in CI. I saw the other resources using it, so copied the pattern. I'd guess it's something to do with the docs not matching the naming convention expected by the bridge.

PR #1057, opened 2023-04-10
Update to v6.54.0 - upstream provider to v4.61.0

kpitzen:


kpitzen:
> What's the deal with the Firebase revert? Is that something to watch out for in future upgrades?

Yes - it, similar to reverts in the previous version, are due to the GCP TF provider moving to plugin framework for some resources / data sources.

thomas11 (approved):

Issue #1051, opened 2023-04-05
[TFPF] Upstream provider is now a mixed SDK/TFPF provider


kpitzen:
### What happened?

As of the most recent release, the `google-beta` provider has begun to include TFPF resources alongside existing TF SDK resources. To support bridging these resources out of the box, we will require the tfbridge muxer work to be applied to this provider as well.

### Expected Behavior

Resources like this: https://github.com/pulumi/terraform-provider-google-beta/blob/patched-v4.60.0/google-beta/data_source_dns_keys.go

Should be bridged into the Pulumi provider as expected, rather than backported as a `reverted` resource: https://github.com/pulumi/terraform-provider-google-beta/blob/patched-v4.60.0/google-beta/data_source_dns_keys_reverted.go

### Steps to reproduce

Run `make tfgen` on a new tag of the upstream provider. Notice that new TFPF resources are missing from the output.

### Output of `pulumi about`

N/A

### Additional context

Existing epic tracking TFPF work in the bridge as well as the Muxer work mentioned above:
https://github.com/pulumi/pulumi-terraform-bridge/issues/744

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

PR #1075, opened 2023-05-01
Enable Plugin Framework based resources and functions for pulumi-gcp

iwahbe:
Fixes https://github.com/pulumi/pulumi-gcp/issues/1051

Most of the diff in the PR comes from upgrading the bridge to the latest that can be done as a separate PR.

Here is also some context on deploying https://github.com/pulumi/pulumi-terraform-bridge/issues/981 providers with a muxed SDKv2/PF design.

This currently does not have any known blockers and smaller providers are now in prod exercising the necessary code.

Per @iwahbe testing we are unable to test the new datasource due to test account limitations.


iwahbe:
This change adds a single datasource: `getAndroidAppConfig`. In an attempt to test the datasource, I tried to deploy an `AndroidApp`. This is the message I got:

```
Diagnostics:
  gcp:firebase:AndroidApp (app):
    error: 1 error occurred:
        * Error creating AndroidApp: googleapi: Error 403: Your application has authenticated using end user credentials from the Google Cloud SDK or Google Cloud Shell which are not supported by the firebase.googleapis.com. We recommend configuring the billing/quota_project setting in gcloud or using a service account through the auth/impersonate_service_account setting. For more information about service accounts and how to use them in your application, see https://cloud.google.com/docs/authentication/. If you are getting this error with curl or similar tools, you may need to specify 'X-Goog-User-Project' HTTP header for quota and billing purposes. For more information regarding 'X-Goog-User-Project' header, please check https://cloud.google.com/apis/docs/system-parameters.
    Details:
    [
      {
        "@type": "type.googleapis.com/google.rpc.ErrorInfo",
        "domain": "googleapis.com",
        "metadata": {
          "consumer": "projects/764086051850",
          "service": "firebase.googleapis.com"
        },
        "reason": "SERVICE_DISABLED"
      }
    ]
```

kpitzen:
Hey @t0yv0 and @iwahbe - thanks for pushing this work forward! Does adding this change the upgrade / release process for the provider re: what's currently in our published playbooks?

t0yv0:
@kpitzen not intentionally. Can you point me to the runbooks I can work with you on that. One obvious change is the new dependency on `pulumi-terraform-bridge/pf` module, that needs to be kept up to date.

t0yv0:
FYI Ian is on PTO until May 30 but I can answer any questions in the meanwhile.

provider/resources.go
```diff
@@ -348,9 +350,13 @@ func preConfigureCallbackWithLogger(ctx context.Context, host *provider.HostClie
 	return nil
 }
 
+//go:embed cmd/pulumi-resource-gcp/bridge-metadata.json
+var metadata []byte
+
 // Provider returns additional overlaid schema and metadata associated with the gcp package.
 func Provider() tfbridge.ProviderInfo {
 	p := shimv2.NewProvider(google.Provider())
+	p = pf.MuxShimWithPF(context.Background(), p, google.New(""))
```
danielrbradley, line 359: What's the significance of the `google.New("")` - why the empty string?

danielrbradley (commented):
Happy with all the implementation.

Let's split this PR so we've got a simple bridge upgrade PR first so we can see if there's any significant SDK changes.

Looking at we've reverted it looks like only data sources have been implemented in tfpf so far - so we're not seeing any added resources?

Once we're ready to merge, I'd suggest also including into this PR reverting framework revert patch from the fork as part of the upgrade so we don't carry it forward

t0yv0:
> Looking at we've reverted it looks like only data sources have been implemented in tfpf so far - so we're not seeing any added resources?

Yes, this is a bit strange; do you know which resources we expect? Is this because some PF resources were poly-filled in a different way? Can you give some pointers?

> Once we're ready to merge, I'd suggest also including into this PR reverting framework revert patch from the fork as part of the upgrade so we don't carry it forward

Any pointers?

provider/resources.go
t0yv0: It's a version. Perhaps version.Version would be more appropriate here, something we could try. 

Looks like the version is used by the pluginFrameworkProvider in particular in responses to GetMetadata function but I doubt we depend on that being accurate.

```go
// New is a helper function to simplify provider server and testing implementation.
func New(version string) provider.ProviderWithMetaSchema {
	return &frameworkProvider{
		version: version,
	}
}
```

t0yv0:
https://github.com/pulumi/pulumi-gcp/pull/1090 baseline bridge update

t0yv0:
Updated the PR. The diff is much smaller and highlights that's only one method is being added.

t0yv0:
Per conv with @mikhailshilkov this provider has not been patching PF-based resources. Checks on upstream code. No PF resources:

```
// Resources defines the resources implemented in the provider.
func (p *frameworkProvider) Resources(_ context.Context) []func() resource.Resource {
	return nil
}
```

We should be seeing more data sources though:

```

// DataSources defines the data sources implemented in the provider.
func (p *frameworkProvider) DataSources(_ context.Context) []func() datasource.DataSource {
	return []func() datasource.DataSource{
		NewGoogleClientConfigDataSource,
		NewGoogleClientOpenIDUserinfoDataSource,
		NewGoogleDnsManagedZoneDataSource,
		NewGoogleDnsRecordSetDataSource,
		NewGoogleDnsKeysDataSource,
		NewGoogleFirebaseAndroidAppConfigDataSource,
		NewGoogleFirebaseAppleAppConfigDataSource,
		NewGoogleFirebaseWebAppConfigDataSource,
	}
}
```

I'l have a closer look as to why we are not seeing those.

t0yv0:
Thanks @danielrbradley for pointing out an upstream commit that patched upstream to serve the missing datasources from SDKv2 versions. I've tried reverting this commit on an upstream branch and building again, hoping this move all the datasources onto PF implementation now.

```
Revert framework conversions
- Leave the new framework versions in-place.
- Rename the reverted versions to have the filename suffix `_reverted` and the function name suffix `...Sdk`.
- Re-insert into old DataSourcesMap in provider.go
```

t0yv0:
So the 7 functions are not new according to schema-tools check but they're switching over to PF implementations as evidenced in the edit to bridge-metadata.json dispatch mappings.

kpitzen (approved):
Apologies for the late reply - agreed that this LGTM

kpitzen:
> Couple more quick questions before I merge.
> 
> Does the [pulumi/terraform-provider-google-beta@e22b067c5b69](https://github.com/pulumi/terraform-provider-google-beta/commit/e22b067c5b69) change look good?
> 
> Should I tag this in the upstream somehow?

This LGTM - we had to do something similar as the upstream TF provider refactored some of the config functionality. Shouldn't be a need to tag upstream.

Issue #1074, opened 2023-04-29
Misspellings for Storage Transfer service account related parts of the code base in all SDKs


lrnq:
### What happened?

I needed to use the method described here: https://github.com/pulumi/pulumi-gcp/blob/143e991628e834142e02ae0043cc5f8dc15cded1/sdk/python/pulumi_gcp/storage/get_transfer_project_servie_account.py#LL20C6-L20C6 to retrieve the Storage Transfer service account for a particular project. 

I noticed that the function name is misspelled. From what I can tell this is the case across all parts of the codebase related to the Storage Transfer service account for all the SDKs. 

Running `grep -inr "servie" sdk | awk -F: '{print $1":"$2" --- "$3}'` yields the following:
```
sdk/nodejs/storage/transferJob.ts:27- * const default = gcp.storage.getTransferProjectServieAccount({
sdk/nodejs/storage/getTransferProjectServieAccount.ts:16- * const default = gcp.storage.getTransferProjectServieAccount({});
sdk/nodejs/storage/getTransferProjectServieAccount.ts:20-export function getTransferProjectServieAccount(args?
sdk/nodejs/storage/getTransferProjectServieAccount.ts:24-    return pulumi.runtime.invoke("gcp
sdk/nodejs/storage/getTransferProjectServieAccount.ts:30- * A collection of arguments for invoking getTransferProjectServieAccount.
sdk/nodejs/storage/getTransferProjectServieAccount.ts:32-export interface GetTransferProjectServieAccountArgs {
sdk/nodejs/storage/getTransferProjectServieAccount.ts:40- * A collection of values returned by getTransferProjectServieAccount.
sdk/nodejs/storage/getTransferProjectServieAccount.ts:42-export interface GetTransferProjectServieAccountResult {
sdk/nodejs/storage/getTransferProjectServieAccount.ts:70- * const default = gcp.storage.getTransferProjectServieAccount({});
sdk/nodejs/storage/getTransferProjectServieAccount.ts:74-export function getTransferProjectServieAccountOutput(args?
sdk/nodejs/storage/getTransferProjectServieAccount.ts:75-    return pulumi.output(args).apply((a
sdk/nodejs/storage/getTransferProjectServieAccount.ts:79- * A collection of arguments for invoking getTransferProjectServieAccount.
sdk/nodejs/storage/getTransferProjectServieAccount.ts:81-export interface GetTransferProjectServieAccountOutputArgs {
[...]

PR #1076, opened 2023-05-01
Update misspelled getTransferProjectServiceAccount function

rquitales:
Fixes: #1074 

This is possibly a breaking change since we rename the following data source:
`getTransferProjectServieAccount` -> `getTransferProjectServiceAccount`


thomas11 (changes requested):
We have `prov.RenameDataSource` to make this a non-breaking change. There are some examples already in resources.go.

thomas11 (approved):

PR #1087, opened 2023-05-16
Update to v4.65.1

kpitzen:
Closes https://github.com/pulumi/pulumi-gcp/issues/1085


danielrbradley:
@kpitzen taking the `serviceDirectoryRegistrations` example - it looks like they're changing it to a single item instead of a list dispite the name still being plural.

If this is the case, then we can revert back to a list by overriding [MaxItemsOne](https://github.com/pulumi/platform-providers-team/blob/main/playbooks/maxitemsone.md)

danielrbradley:
Superseded by:
- #1088

PR #1096, opened 2023-06-06
Add AutoAliasing

iwahbe:
Adds auto-aliasing for resources and max-items-one retention.

### DIFF
--- a/provider/cmd/pulumi-resource-gcp/bridge-metadata.json
+++ b/provider/cmd/pulumi-resource-gcp/bridge-metadata.json
@@ -2624,6 +2624,9 @@
                                                 "maxItemsOne": false,
                                                 "elem": {
                                                     "fields": {
+                                                        "empty_dir": {
+                                                            "maxItemsOne": true
+                                                        },
                                                         "secret": {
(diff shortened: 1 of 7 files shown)


rquitales (commented):
This change looks fine, though I'm lacking context about whether this will have the intended behaviour? Should we add some tests for this to better explain what this change does?

iwahbe:
> This change looks fine, though I'm lacking context about whether this will have the intended behaviour? Should we add some tests for this to better explain what this change does?

The intended behavior is to maintain new aliases and MaxItemsOne values until the next major version. I have removed a manual alias and all MaxItemsOne values currently set,  ensuring that we rely on `AutoAliasing` to avoid a semantic diff in the schema.

Since we see a very small cosmetic diff in the schema (only in 1 resource's deprecation message), we know that `AutoAliasing` is working correctly.

We don't need to add tests for `AutoAliasing` per se. It is well tested in the bridge. If it breaks for gcp, that will surface as spurious schema diffs and will not be seen by our users.

rquitales (approved):
Thanks for the explanation!

PR #1103, opened 2023-06-20
Replace upstream fork with submodule

### LINKED ISSUE #1102: Migrate fork to upstream submodule
This would enable the upgrade provider action to manage this provider as any others by the ecosystem team, e.g.: pulumi/pulumi-azure.

aq17:
Fixes #1102 

Since team ecosystem now owns `pulumi-gcp`, migrating from maintaining a fork to using the submodule aligns this provider with the rest of ecosystem's providers, allowing us to utilize the `upgrade-provider` tool on it.

### DIFF
--- a/patches/0001-fork.patch
+++ b/patches/0001-fork.patch
@@ -1,0 +1,6220 @@
+diff --git b/CHANGELOG.md a/CHANGELOG.md
+index 5131ff6ab..f4340b2fc 100644
+--- b/CHANGELOG.md
++++ a/CHANGELOG.md
+@@ -66,6 +66,32 @@ BUG FIXES:
+ * provider: fixed an issue where mtls transports were not used consistently(initial implementation in v4.65.0, reverted in v4.65.1) ([#5645](https://github.com/hashicorp/terraform-provider-google-beta/pull/5645))
+ * storage: fixed inconsistent final plan when labels are added to `google_storage_bucket` ([#5634](https://github.com/hashicorp/terraform-provider-google-beta/pull/5634))
+ 
++NOTE:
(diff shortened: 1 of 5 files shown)


patches/0001-fork.patch
```diff
@@ -0,0 +1,6220 @@
+diff --git b/CHANGELOG.md a/CHANGELOG.md
+index 5131ff6ab..f4340b2fc 100644
+--- b/CHANGELOG.md
++++ a/CHANGELOG.md
+@@ -66,6 +66,32 @@ BUG FIXES:
+ * provider: fixed an issue where mtls transports were not used consistently(initial implementation in v4.65.0, reverted in v4.65.1) ([#5645](https://github.com/hashicorp/terraform-provider-google-beta/pull/5645))
+ * storage: fixed inconsistent final plan when labels are added to `google_storage_bucket` ([#5634](https://github.com/hashicorp/terraform-provider-google-beta/pull/5634))
+ 
++NOTE:
```
AaronFriel, line 9: This seems off, doesn't look like code we would have added.

AaronFriel (commented):
Are there files/lines in the patch that were actually contributed upstream or for some reason weren't in our diff?

aq17:
@AaronFriel I created the patch using the diff of [this branch](https://github.com/pulumi/terraform-provider-google-beta/tree/patched-v4.68.0), which it looks like is the most recent one @thomas11 created when upgrading

AaronFriel:
Interesting - I see that diff. Makes sense, we can correct the patch if it poses us issues down the road.

AaronFriel (approved):

PR #1180, opened 2023-09-06
Standardize capitalization for `Iam` (from `IAM`)

iwahbe:
Standardize the capitalization of iam to `Iam` (not `IAM`). This is currently the most commonly used capitalization. 

~Despite what our schema changes says, this is actually a no-op for every language. Pulumi YAML should continue to work as expected, since it [respects aliases](https://github.com/pulumi/pulumi-yaml/blob/0db26f765692a4aef1c339b1b3dd2326d74071b6/pkg/pulumiyaml/README.md#L15-L16).~


provider/cmd/pulumi-resource-gcp/bridge-metadata.json
```diff
@@ -26114,6 +26115,7 @@
             "gcp:organizations/getFolder:getFolder": 0,
             "gcp:organizations/getFolders:getFolders": 0,
             "gcp:organizations/getIAMPolicy:getIAMPolicy": 0,
+            "gcp:organizations/getIamPolicy:getIamPolicy": 0,
```
t0yv0, line 26118: I feel worried about these two side-by-side entries that differ only in casing.

```
          "gcp:organizations/getIAMPolicy:getIAMPolicy": 0,
            "gcp:organizations/getIamPolicy:getIamPolicy": 0,
```

Are these really distinct tokens?

iwahbe: I've changed the purpose of this PR to resolve this issue. I'll open a new PR to remove the patches.

t0yv0:
I'm very sceptial about the assertion this is a no-op change. 

At the very least the docsgen changes indicate that there's either a bug being fixed or a bug being introduced in programgen, or else we now generate double the classes in the SDK:

```
  var iam = new Gcp.Kms.CryptoKeyIAMMember("iam", new()
```

I'd also worry about out of band things like registry URLs or registry site renders if they key on tokens. 

Given the risks involved, what are the benefits? What is driving this work?

t0yv0:
This has been stalling a bit. Further plans here? You believe aliasing will make this "essentially" non-breaking?

The type changes are breaking but perhaps justifiable and we need to call it out in rel notes:

"gcp:compute/InstanceIAMBindingCondition:InstanceIAMBindingCondition" missing

iwahbe:
I'm afraid that we will need to table this until v8.

PR #1184, opened 2023-09-08
Patch Cloud SQL Database Instance to refresh settings.version

### LINKED ISSUE #794: Error 412 while updating gcp.sql.DatabaseInstance
### What happened?

I'm attempting to enable the High Availability setting for an existing GCP Postgres SQL instance. Adding the availability type parameter returns the following error 

```
  pulumi:pulumi:Stack (app):
    error: update failed
 
  gcp:sql:DatabaseInstance (db-server):
    error: 1 error occurred:
    	* updating urn:pulumi:test::app::gcp:sql/databaseInstance:DatabaseInstance::db-server: 1 error occurred:
    	* Error, failed to update instance settings for : googleapi: Error 412: Condition does not match., staleData
```

### Steps to reproduce

Create a DB instance

```
export const dbInstance = new gcp.sql.DatabaseInstance("db-server", {
  databaseVersion: "POSTGRES_11",
  region: "us-central1",
  settings: {
    tier: "db-g1-small",
    diskAutoresize: false,
    backupConfiguration: {
      enabled: false,
      location: "us-central1",
      pointInTimeRecoveryEnabled: false,
      startTime: "01:00",
      transactionLogRetentionDays: 7,
      backupRetentionSettings: {
        retainedBackups: 14,
      },
    },
    userLabels: {
      "env": "test",
    }
  },
  deletionProtection: false,
});
```

Add `availabilityType` to the instance settings

```
export const dbInstance = new gcp.sql.DatabaseInstance("db-server", {
  databaseVersion: "POSTGRES_11",
  region: "us-central1",
  settings: {
    tier: "db-g1-small",
    diskAutoresize: isProduction,
    availabilityType: "REGIONAL",
    backupConfiguration: {
      enabled: false,
      location: "us-central1",
      pointInTimeRecoveryEnabled: false,
      startTime: "01:00",
      transactionLogRetentionDays: 7,
      backupRetentionSettings: {
        retainedBackups: 14,
      },
    },
    userLabels: {
      "env": "test",
    }
  },
  deletionProtection: false,
```

Run `pulumi up`

### Expected Behavior

The existing GCP SQL database instance availability is changed from `ZONAL` to `REGIONAL`.

### Actual Behavior

The update fails and returns the error above. The following data was also in the logs: 
```
I0416 18:13:19.856905     770 eventsink.go:59] {
I0416 18:13:19.857036     770 eventsink.go:59]   "error": {
I0416 18:13:19.857204     770 eventsink.go:59]     "code": 412,
I0416 18:13:19.857382     770 eventsink.go:59]     "message": "Condition does not match.",
I0416 18:13:19.857524     770 eventsink.go:59]     "errors": [
I0416 18:13:19.857688     770 eventsink.go:59]       {
I0416 18:13:19.857862     770 eventsink.go:59]         "message": "Condition does not match.",
I0416 18:13:19.858078     770 eventsink.go:59]         "domain": "global",
I0416 18:13:19.858259     770 eventsink.go:59]         "reason": "staleData",
I0416 18:13:19.858431     770 eventsink.go:59]         "location": "If-Match",
I0416 18:13:19.858601     770 eventsink.go:59]         "locationType": "header"
I0416 18:13:19.858765     770 eventsink.go:59]       }
I0416 18:13:19.858935     770 eventsink.go:59]     ]
I0416 18:13:19.859112     770 eventsink.go:59]   }
I0416 18:13:19.859292     770 eventsink.go:59] }
```

### Versions used

Pulumi version: 3.29.1

### Additional context

Adding anything appears to trigger the error, including explicitly entering the default value `availabilityType: "ZONAL"`

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

AaronFriel:
Fixes #794, see also upstream issue:
- https://github.com/hashicorp/terraform-provider-google/issues/15762

Cloud SQL Database Instance resources fail to update in Pulumi when the resource property `settings.settingsVersion` is modified outside of the Pulumi program. The API specification requires updates set that property to the current value of the field. When the value drifts outside of Pulumi's control, the mismatch results in error 412 on PATCH.

See: https://cloud.google.com/sql/docs/mysql/admin-api/rest/v1/instances#settings

This modifies the resource to always read the current value before performing the PATCH request that updates the resource.

Testing this posed a few challenges for both writing an upstream acceptance test and for our own test framework.

For the upstream, the TF SDKv2 test harness has a mandatory `refresh` operation that runs before applying a configuration. For the upstream issue, a script was written against the `terraform` CLI to simulate drift, published at [AaronFriel/google-cloud-sql-instance-412-bug](https://github.com/AaronFriel/google-cloud-sql-instance-412-bug). That script creates a resource, imports it into a second state and modifies it, then attempts to update the original state again.

For our own acceptance tests, the challenge was that we lack prior art in this repository for calling Google Cloud APIs directly. Lacking an API to call to obtain Google Cloud credentials in the test harness (in the same manner as the provider does), this PR instead uses the automation API with inline programs to perform the same set of steps as the test script above.


t0yv0:
At low level LGTM, thank you for this contribution, needs go mod tidy. I think we should take the patch @pulumi/providers 

At a high level, what is the reason that Pulumi providers do not refresh on update by default? There is a possibility that upstream issue will never be addressed because it works as expected with the TF model. Could the case be made for these alternatives:

1. bridged providers start refresh on update wholesale, to match TF behavior

2. bridged providers get a per-resource switch to enable refresh on update; this is essentially what the patch does, for this one resource; having a ProviderInfo switch would make it ergonomic to do a lot more of this kind of modification without patching upstream

3. should we reject this as "works by design", you have drift so now Pulumi expects you to opt into refresh explicitly? 

Personally I'm not really for option 3 but listing it here for completeness.

t0yv0 (approved):

PR #1216, opened 2023-09-26
Add project_name for goreleaser

rquitales:
When CI is upgraded to goreleaser v1.21, our publish steps fail as the `project_name` could not be formulated by GoReleaser. This PR explicitly sets the `project_name`.

Note, for faster feedback loop, this PR should be merged first and a follow-up PR in pu/ci-mgmt will be created.
I've also triggered the CI workflow for this commit to test that it works and it is passing: https://github.com/pulumi/pulumi-gcp/actions/runs/6313530404/job/17142821989

Fixes: #1217

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -441,7 +441,6 @@ on:
   push:
     branches:
     - master
-    - rquitales/fix-goreleaser
     paths-ignore:
     - "**.md"
     tags-ignore:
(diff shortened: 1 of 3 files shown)


thomas11 (commented):
GCP is managed by ci-mgmt. Can we not make this check there, centrally?

t0yv0 (approved):

t0yv0:
Guess we're updating ci-mgmt afterward.

Issue #1148, opened 2023-06-27
resource names should not use camelCase as it causes errors, a kebab-case would be a solution for this


Jaafar-Nasrallah:
[themes/default/content/registry/packages/gcp/api-docs/compute/targethttpproxy/_index.md](https://www.pulumi.com/registry/packages/gcp/api-docs/compute/targethttpproxy/)

in the examples. resources where created with strings such as `defaultTargetHttpProxy` and `defaultURLMap` and these names produce an error
a solution would be to use `default-target-http-proxy` and `default-url-map` instead

lukehoban:
This is an issue with the default autonaming being applied to this resource.  You can workaround this by passing `name: ` explicitly.  But we should also fix this by adding an override similar to https://github.com/pulumi/pulumi-gcp/blob/41c8c3bc39a2cdfc927808d1074a3cf7750bedaf/provider/resources.go#L771 for this resource.

Issue #813, opened 2022-05-19
Pulumi generated resource name for `gcp.compute.HttpHealthCheck()` doesn't comply with the provider's expected name regex


aureq:
### What happened?

With the example below (taken from our [doc](https://www.pulumi.com/registry/packages/gcp/api-docs/compute/backendservice/)), it appears the Pulumi generated resource name for `gcp.compute.HttpHealthCheck()` doesn't comply with the provider's expected name regex.

The preview phase works as expected and no error is shown. However, during deployment an error is returned by the Google API.

```
Error creating HttpHealthCheck: googleapi: Error 400: Invalid value for field 'resource.name': 'defaultHttpHealthCheck-886eaee'. Must be a match of regex '(?:[a-z](?:[-a-z0-9]{0,61}[a-z0-9])?)', invalid
```

However, setting the _cloud_ resource name manually to match the regex allows the deployment to succeed.

### Steps to reproduce

1. Deploy the app below

```ts
import * as pulumi from "@pulumi/pulumi";
import * as gcp from "@pulumi/gcp";

export = async () => {

    const defaultHttpHealthCheck = new gcp.compute.HttpHealthCheck("defaultHttpHealthCheck", {
        // name: "abs123",
        requestPath: "/",
        checkIntervalSec: 1,
        timeoutSec: 1,
    });
   return {

    }
}


```

### Expected Behavior

The deployment succeeds.

### Actual Behavior

The deployment fails during the `up` phase and Google API returns the error below.

```
Error creating HttpHealthCheck: googleapi: Error 400: Invalid value for field 'resource.name': 'defaultHttpHealthCheck-886eaee'. Must be a match of regex '(?:[a-z](?:[-a-z0-9]{0,61}[a-z0-9])?)', invalid
```

### Versions used

```
$ pulumi about
CLI          
Version      3.32.1
Go Version   go1.17.9
Go Compiler  gc

Plugins
NAME    VERSION
gcp     6.24.0
nodejs  unknown

Host     
OS       debian
Version  11.3
Arch     x86_64

This project is written in nodejs (/usr/local/bin/node v16.15.0)

Current Stack: dev

TYPE                                         URN
pulumi:pulumi:Stack                          urn:pulumi:dev::gcp-ts-backend-service::pulumi:pulumi:Stack::gcp-ts-backend-service-dev
[...]

PR #1235, opened 2023-10-06
Refine auto-naming for a few HTTP-based resources

mikhailshilkov:
We have a few reports about auto-naming failing when a given resource type does not allow capital letters in it. The motivation seems to be that those names are used as parts of URLs.

The PR overrides auto naming to use lowercase letters for a set of reported resources.

I added a test for one of the resources - the test fails without the fix and passes with the fix.

Fix https://github.com/pulumi/pulumi-gcp/issues/1148
Fix https://github.com/pulumi/pulumi-gcp/issues/813
Fix https://github.com/pulumi/pulumi-gcp/issues/1254

### DIFF
--- a/examples/examples_yaml_test.go
+++ b/examples/examples_yaml_test.go
@@ -1,0 +1,25 @@
+// Copyright 2016-2023, Pulumi Corporation.  All rights reserved.
+//go:build go || all
+// +build go all
+
+package examples
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/pulumi/pulumi/pkg/v3/testing/integration"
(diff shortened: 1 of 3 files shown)


t0yv0:
Could we add network in here ? https://github.com/pulumi/pulumi-gcp/issues/1254 this one of the top resources.

mikhailshilkov:
> Could we add network in here ? https://github.com/pulumi/pulumi-gcp/issues/1254 this one of the top resources.

Great suggestion. Will do tomorrow!

mikhailshilkov:
@t0yv0 I added the Network

rquitales (approved):
LGTM - thanks for the fix!

PR #1263, opened 2023-10-11
Move to setup-dotnet v3

mikhailshilkov:
Will be rolled out as part of https://github.com/pulumi/ci-mgmt/pull/669


danielrbradley (approved):
Should we do this via ci-mgmt or doing it here quickly first?

guineveresaenger:
@danielrbradley - doing it here first serves as a canary to see if the behavior is as expected. It turned out we did catch a missing update on pulumi-package-publisher when we merged to default, so while this is slow and manual, it saved us from trouble this time around.

PR #1264, opened 2023-10-11
Updates to upgrade tests

t0yv0:
Continuing to evolve providertest library; this is introducing the preferred pattern from that library, breaking out top-level tests by test-program and combining upgrades with other forms of testing (e2e, sdk).

Curious about CI results here; this may need a little more support from the Makefile and CI system since these tests run in the `prerequisites` phase which may not have the necessary credentials available to test `pulumi preview`.


t0yv0:
Blocked by https://github.com/pulumi/ci-mgmt/issues/675

t0yv0:
This is now upgraded to 0.0.2 of the provider and passing locally for me. I think there still is a problem with running these in CI. Perhaps I can find a way to skip them in CI until we can circle back to reintroducing creds into the prerequisites job.

t0yv0:
I've worked around the issues with lack of credentials. We can check this in, I think it will help with having the upgrade tests aligned with the latest shape of providertest. CC @guineveresaenger

provider/provider_test.go
```diff
@@ -0,0 +1,148 @@
...
+			t.Skipf("Skip recording baseline behavior because %s", flags.Snapshot.WhyNotSet())
+		}
+	})
+	for _, m := range providertest.UpgradeTestModes() {
+		t.Run(fmt.Sprintf("upgrade-%s", m), func(t *testing.T) {
+			t.Helper()
+			pt.VerifyUpgrade(t, m)
+		})
+	}
+}
```
mjeffryes, line 132: I thought that one of the features of the providertest API was supposed to be that we can easily choose which kinds of tests to run? Can we not just set some flags to only run upgrade tests?

mjeffryes (approved):
Seems like a reasonable step to unblock the state of testing in GCP

provider/provider_test.go
t0yv0: Yeah, we missed making this flag settable at Go level.

guineveresaenger (approved):
Happy to try this.

Thank you for these changes!

PR #1292, opened 2023-10-25
Pulumi tags

iwahbe:
This is a simple recursive name override on `"terraform_labels"` to `"pulumiLabels"`.


provider/tags.go
```diff
@@ -0,0 +1,70 @@
+// Copyright 2016-2018, Pulumi Corporation.  All rights reserved.
```
guineveresaenger, line 1:
```suggestion
// Copyright 2016-2023, Pulumi Corporation.  All rights reserved.
```

provider/tags.go
guineveresaenger, line 1: Since these metadata thingies are called Labels in GCP, might make sense to keep the filename in sync with that, especially since these words are so overloaded.

Really appreciate that this sits in its own file.

provider/tags.go
```diff
@@ -0,0 +1,70 @@
...
+
+package gcp
+
+import (
+	"github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfbridge"
+	shim "github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfshim"
+)
+
+func fixTagNames(prov *tfbridge.ProviderInfo) {
+	// Recursively applies the tag fix
```
guineveresaenger, line 11: I think we may want a couple words here to indicate what the tag/label fix is and why we're doing it.

provider/tags.go
```diff
@@ -0,0 +1,70 @@
+// Copyright 2016-2018, Pulumi Corporation.  All rights reserved.
+
+package gcp
+
+import (
+	"github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfbridge"
+	shim "github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfshim"
+)
+
+func fixTagNames(prov *tfbridge.ProviderInfo) {
```
guineveresaenger, line 10: Can we add a unit test for this functionality?

guineveresaenger (approved):
A couple nits and if we can add a unit test that would be dope.

Issue #1259, opened 2023-10-10
Add PreviewOnly mode coverage for provider upgrade tests


t0yv0:
https://github.com/pulumi/pulumi-gcp/pull/1237 introduces provider upgrade tests but the first version only runs the quick mode. An additional PreviewOnly mode is available that is higher fidelity but is slower to execute. The task is to make sure CI runs that as well, possibly under tests* targets and not prerequisites since it would require cloud credentials.

t0yv0:
With quick-mode tests skipped because of https://github.com/pulumi/providertest this is becoming important.

One possible way forward here would be to add running these tests per language in the test matrix similar to how AWSX did that https://github.com/pulumi/pulumi-awsx/commit/6e0553dc5b158d869ca33ddfc52aaf25c83dfef8

PR #1303, opened 2023-10-31
Ensure CI runs PreviewOnly upgrade tests

t0yv0:
Fixes https://github.com/pulumi/pulumi-gcp/issues/1259

Upgrade tests in PreviewOnly mode, as well as E2E tests over the same test programs, now run in CI. See [example](https://github.com/pulumi/pulumi-gcp/actions/runs/6727319213/job/18285119414?pr=1303#step:25:119).

- This is accomplished by using .ci-mgmt.yml and `make ci-mgmt` to extend the test matrix to run ./provider tests first, before it runs ./examples tests. The test matrix is split by language, therefore the provider tests are now also split, yaml and java languages are currently grouped into one bucket. 

- The tests are run only in the test matrix, and early tests are skipped in prerequisites - as currently no tests can succeed without Google credentials, and prerequisites is still lacking those credentials 

- Negated build tags are used to split the tests in CI while still making them run locally without setting build tags

- Baselines are re-recorded on CI using the CI-native region, project and zone

- To enable running PreviewOnly tests locally, ignoreChanges workaround is added to every program to disregard differences in region, project, and zone between CI-recorded baselines and local pulumi-development project


.ci-mgmt.yaml
```diff
@@ -14,6 +14,7 @@ env:
   GOOGLE_REGION: "us-central1"
   GOOGLE_ZONE: "us-central1-a"
   PULUMI_MISSING_DOCS_ERROR: true
+  VERSION_PREFIX: 7.0.0
```
guineveresaenger, line 17: do we need this here? I think line 3 takes care of it?

provider/provider_test.go
```diff
@@ -4,134 +4,56 @@ package gcp
...
-		{"bigquery-routine-1"},
-		{"bigquery-routine-2"},
-		{"cloudrunv2-job-6"},
-		{"cloudrunv2-service-4"},
-		{"cloudrunv2-service-5"},
-		{"cloudrunv2-service-7"},
-		{"compute-disk-1"},
-		{"compute-disk-3"},
+	defaultValues := map[string]string{
+		"gcp:config:project": "pulumi-development",
```
guineveresaenger, line 30: This is available via env var as well: `GCP_PROJECT_ID`.

guineveresaenger (approved):
My one question is whether this would `ignoreChanges` in CI as well.

.ci-mgmt.yaml
t0yv0: I think we do. Without this entry VERSION_PREFIX env var will not populate. Confusing pulumictl and versioning.

provider/provider_test.go
t0yv0: This is certainly skewing to Pulumi-internal setup but I find it convenient.

t0yv0:
> whether this would ignoreChanges in CI as well.

Yes. That is believed to be benign.

PR #1304, opened 2023-10-31
Try using a bigger host for the prereq build on nightly to see if it resolves the OOM like build failure

mjeffryes:
Should resolve [Workflow failure: cron· pulumi-gcp/1301](https://github.com/pulumi/pulumi-gcp/issues/1301) Eventually, we need to restructure builds to make them more manageable.

### DIFF
--- a/.ci-mgmt.yaml
+++ b/.ci-mgmt.yaml
@@ -33,4 +33,5 @@ skipGoReleaserHooks: true
 javaGenVersion: v0.9.7
 runner:
   publish: pulumi-ubuntu-8core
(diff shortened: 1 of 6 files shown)


iwahbe (approved):
The change is fine, but it will be overwritten by ci-mgmt.

To change this permanently, you will need to set `runner.prerequisites` in `.ci-mgmt.yaml`:

https://github.com/pulumi/ci-mgmt/blob/a50fdc419e81aa956b3665bd0a802f04aeda5237/provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/prerequisites.splice#L1

Issue #722, opened 2022-01-18
Golang sdk, serviceAccount package not exists, should be "serviceaccount". 


Mistic92:
I'm trying to follow [docs ](https://www.pulumi.com/registry/packages/gcp/api-docs/serviceaccount/account/)but I can't use given package and it's not available in package repository

When trying to use  serviceaccount package I'm getting
```
github.com/pulumi/pulumi-gcp/sdk/v6/go/gcp/serviceAccount: module github.com/pulumi/pulumi-gcp/sdk/v6@latest found (v6.7.0), but does not contain package github.com/pulumi/pulumi-gcp/sdk/v6/go/gcp/serviceAccount
```

https://www.pulumi.com/registry/packages/gcp/api-docs/serviceaccount/account/

Also it's not available there
https://pkg.go.dev/github.com/pulumi/pulumi-gcp/sdk/v6@v6.5.0/go/gcp/serviceAccount

Looks like bug in tutorial as correct name is `serviceaccount`
https://pkg.go.dev/github.com/pulumi/pulumi-gcp/sdk/v6@v6.5.0/go/gcp/serviceaccount

salineselin:
I ran into this issue as well when I imported a serviceaccount. the outputted code declared the go module as `serviceAccount` instead of `serviceaccount`

burmecia:
Got same issue and the example on this page https://www.pulumi.com/registry/packages/gcp/api-docs/serviceaccount/account/ is also wrong:

```
import * as pulumi from "@pulumi/pulumi";
import * as gcp from "@pulumi/gcp";

const serviceAccount = new gcp.serviceAccount.Account("service_account", {
    accountId: "service-account-id",
    displayName: "Service Account",
});
```

It should be:

```
const serviceAccount = new gcp.serviceaccount.Account("service_account", {
    accountId: "service-account-id",
    displayName: "Service Account",
});
```

shadiramadan:
Seriously still not fixed... The schema.json is full of issues.  I found a few typos for some other resources as well.

anilkumarnm:
Still not fixed :-(

PR #1305, opened 2023-11-01
Rename serviceAccount to serviceaccount

mikhailshilkov:
Renames the `serviceAccount` module to `serviceaccount` according to our convention of lowercase modules. It's been a long-standing override that we can now get rid of. The change adds aliases to the old type names. In terms of SDKs, this change is a non-breaking and should have no direct user impact.

Once v7 ships, this will fix https://github.com/pulumi/pulumi-gcp/issues/722


iwahbe (approved):
LGTM

provider/resources.go
```diff
@@ -3883,16 +3873,14 @@ func Provider() tfbridge.ProviderInfo {
 			},
 		})
 
-	err := x.ComputeDefaults(&prov, x.TokensMappedModules("google_", "",
+	prov.MustComputeTokens(tks.MappedModules("google_", "",
```
guineveresaenger, line 3876: Thank you for modernizing token mapping! 🙏

guineveresaenger (approved):

PR #1308, opened 2023-11-01
Fix nested tf labels

iwahbe:
The previous fix didn't work correctly for nested labels. This fix *does* work, and it adds a direct assert on the schema that unwanted names do not show up.


guineveresaenger (approved):
LGTM

PR #1238, opened 2023-10-06
[DO NOT MERGE] Upgrade terraform provider google beta to v5.0.0

guineveresaenger:
This pull request presents a major version upgrade for the Pulumi Google Provider. This work is tracked in #1183.

- Removes deprecated resources and hard aliases; see schema-diff summary below.
- Removes security policy patch as fixed via [the bridge](https://github.com/pulumi/pulumi-gcp/issues/1029#issuecomment-1472632119) and adds integration test to verify it works as expected
- Addresses [bigquery.dataset patch](https://github.com/pulumi/pulumi-gcp/issues/1187) as still needed and adds test to verify correct behavior
- Adjusts patch set for upstream changes
- Tests remain enabled for major version upgrade 🎉 
- Fixes `bucket-go` tests

Ready for review; waiting on [bridge update](https://github.com/pulumi/pulumi-gcp/pull/1256) and [upgrade tests](https://github.com/pulumi/pulumi-gcp/pull/1237) to merge and rebase.

Recommendation: When merged, cut alpha release and pivot to #1248.

Fixes #1229
Fixes #1189
Fixes #1187

PR #1309, opened 2023-11-01
Adds a labels test matrix

t0yv0:
On top of https://github.com/pulumi/pulumi-gcp/pull/1238

The new release of GCP adds a DefaultLabels feature.

This change ports a test matrix from a similar feature of AWS provider. The matrix tests that for
all interesting label combinations (empty, non-empty, 0, 1, or two labels, at provider DefaultLabels
or resource Labels level), Pulumi provider can transition a sample resource from one combination to
another, and the actual effective set of labels is what you would expect.

What is currently not tested here but may need scrutiny based on similar issues in AWS:

- passing unknowns into Labels or DefaultLabels
- passing secrets into Labels and making sure this avoids secrets leaks
- EffectiveLabels feature of the GCP provider that detects drift and pulls labels from the cloud
  into management under GCP; this is not tested here
- working with labels in Plugin Framework based resources
  
The tests are currently failing likely due to a bridge issue:
https://github.com/pulumi/pulumi-terraform-bridge/issues/1501


examples/examples_go_test.go
```diff
@@ -1,15 +1,15 @@
 // Copyright 2016-2017, Pulumi Corporation.  All rights reserved.
-//go:build go || all
-// +build go all
```
iwahbe, line 3: I think we need to put these back or our go tests will run for each language.

guineveresaenger (commented):
Let's pull in the bridge commit here so we show the tests will pass.

PR #1310, opened 2023-11-01
T0yv0/add labels test matrix

iwahbe:
This is a reviewable copy of https://github.com/pulumi/pulumi-gcp/pull/1309.


examples/examples_go_test.go
```diff
@@ -19,6 +18,8 @@ import (
 	"github.com/pulumi/pulumi/sdk/v3/go/common/workspace"
 	"github.com/pulumi/pulumi/sdk/v3/go/pulumi"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"os"
```
iwahbe, line 22: `"os"` should be in the top block of imports.

t0yv0:
Feel free to edit my PR as in change base etc.

examples/examples_go_test.go
```diff
@@ -179,3 +180,178 @@ func GenerateRandomString(n int) (string, error) {
...
+				Labels:        label2,
+			})
+		}
+	}
+
+	t.Logf("total state space: %v states", len(states))
+	t.Logf("random-sampling 100 state transitions")
+
+	for i := 0; i < 100; i++ {
+		t.Run(fmt.Sprintf("test-%d", i), func(t *testing.T) {
```
iwahbe, line 263: `""` is interpreted as `fmt.Sprintf("#%d", i)`
```suggestion
		t.Run("", func(t *testing.T) {
```

iwahbe (commented):
2 style nits, but otherwise this looks great.

examples/examples_go_test.go
t0yv0: Have you automated ordering imports yet? I use C-c C-a for go-import-add but it does not do the right thing. On save I use gofmt. Perhaps there's something stronger than gofmt that could reorder them the way we need to on save?

iwahbe: I have automated sorting them within blocks, but I have not automated sorting between blocks. I generally add imports manually, sometimes `eglot` does automatic insertion. I run `gofmt` on save, but that only alphabetizes, not sorting into blocks.

iwahbe: https://github.com/mvdan/gofumpt is the stronger version. You could try `(setq gofmt-command "gofumpt")`.

Issue #1311, opened 2023-11-02
TestRegression794 possibly leaking an expensive DB instance in the test account


t0yv0:
### What happened?

Per https://github.com/pulumi/gcp-account-cleanup/issues/21 these are auto-cleaned up but the test is still creating instances:

```
[
  {
    "name": "test-regression-794-069fea6b",
    "createTime": "2023-11-01T21:19:51.992Z"
  },
  {
    "name": "test-regression-794-ac4ed450",
    "createTime": "2023-11-01T23:57:17.116Z"
  },
  {
    "name": "test-regression-794-e937e342",
    "createTime": "2023-11-01T23:56:58.468Z"
  },
  {
    "name": "test-regression-794-f925ef9b",
    "createTime": "2023-11-01T18:55:40.790Z"
  },
  {
    "name": "test-regression-794-a3e6c40b",
    "createTime": "2023-11-01T23:57:07.330Z"
  },
  {
    "name": "test-regression-794-a57a89e1",
    "createTime": "2023-11-01T23:56:35.194Z"
  },
  {
    "name": "test-regression-794-8ee4c45e",
    "createTime": "2023-11-01T23:57:11.678Z"
  }
]
```

Removing them immediately would further help cost control.

### Example

N/A

### Output of `pulumi about`

N/A

### Additional context

N/A

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

rquitales:
Looks like a couple issues here:

1. `DeletionProtection: pulumi.BoolPtr(false),` needs to be explicitly set when creating the DB Instance as the default is true (https://www.pulumi.com/registry/packages/gcp/api-docs/sql/databaseinstance/)
2. The cleanup function is destroying an empty stack, so the actual DB instance isn't being deleted - needs more investigation

PR #1312, opened 2023-11-02
Fix test cleanup for TestRegression794

rquitales:
This PR fixes the clean up steps for `TestRegression794` to prevent leaking an expensive Cloud SQL DB instance. The 2 changes required are:

1. `DeletionProtection` needs to explicitly be set to `false`
2. The stack backend used for deletion by default attempts to use Pulumi Cloud. This causes an empty stack to be used for running `pulumi destroy`, which will not destroy the DB instance. We need to also specify the `PULUMI_BACKEND_URL` to point to the local backend we've been using in prior steps

Fixes: #1311

### DIFF
--- a/examples/examples_go_test.go
+++ b/examples/examples_go_test.go
@@ -78,7 +78,10 @@ func TestRegression794(t *testing.T) {
 
 	t.Cleanup(func() {
 		// Select an empty program to delete the stack.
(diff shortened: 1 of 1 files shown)


rquitales:
Confirmed that the test SQL Instance created in the presubmit test for this PR cleaned up after itself successfully.

t0yv0 (approved):

Issue #1314, opened 2023-11-03
Persistent Diff on upgrade caused by `pulumiLabels` field introduced via custom diff


guineveresaenger:
During upgrade tests for v6 -> v7, we discovered the following situation:

- If a resource has a `labels` field, it will receive two new output fields as part of the upgrade: `pulumiLabels` and `effectiveLabels`, via provider custom diff.
- When upgrading, the new outputs will result in a no-op diff. This is expected.
- However, after the upgrade, the `pulumiLabels` field remains nil in the pulumi state due to an upstream bug. 
- This means that every time we attempt to upgrade, Pulumi will see a diff on the output value, creating a permanent diff
- In Terraform, this bug gets clobbered by a default Refresh, and on our end, if we run and update using the `--refresh` option, this issue resolves. However, this is poor user experience.

To fix, we will add a runtime state transformation on GCP v7 that ensures that the `pulumiLabels` field always exists
- check if a `pulumiLabels` field exists at the expected path
- if it doesn't exist, set the field to an empty property map

PR #1315, opened 2023-11-03
Fix permanent diff on Pulumi Labels via state transform function

guineveresaenger:
Addresses #1314.

### DIFF
--- a/provider/labels.go
+++ b/provider/labels.go
@@ -3,8 +3,10 @@
 package gcp
 
 import (
+	"context"
 	"github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfbridge"
 	shim "github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfshim"
+	"github.com/pulumi/pulumi/sdk/v3/go/common/resource"
 )
 
 // GCP has many resources with labels called "terraform_labels". We want to change those
@@ -15,10 +17,16 @@ import (
 // tfbridge.SchemaInfo.
 func fixLabelNames(prov *tfbridge.ProviderInfo) {
 	// Recursively applies the label fix
-	var apply func(string, shim.Schema, *tfbridge.SchemaInfo)
-	apply = func(key string, m shim.Schema, info *tfbridge.SchemaInfo) {
+	var apply func(string, shim.Schema, *tfbridge.SchemaInfo, *[]resource.PropertyPath, resource.PropertyPath)
+
+	apply = func(key string, m shim.Schema, info *tfbridge.SchemaInfo,
+		paths *[]resource.PropertyPath,
+		currentPath resource.PropertyPath,
+	) {
 		if key == "terraform_labels" {
 			info.Name = "pulumiLabels"
+			// Gather all paths to pulumiLabels
+			*paths = append(*paths, currentPath)
 		}
 
 		if m == nil {
@@ -26,7 +34,7 @@ func fixLabelNames(prov *tfbridge.ProviderInfo) {
 		}
 
 		switch elem := m.Elem().(type) {
-		case shim.Resource:
+		case shim.Resource: // this is a nested object
 			obj := elem.Schema()
 			if info.Elem == nil {
 				info.Elem = new(tfbridge.SchemaInfo)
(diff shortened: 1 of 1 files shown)


provider/labels.go
```diff
@@ -3,8 +3,10 @@
 package gcp
 
 import (
+	"context"
```
iwahbe, line 6: Std lib should be a separate import block.
```suggestion
	"context"
	
```

provider/labels.go
```diff
@@ -15,18 +17,24 @@ import (
 // tfbridge.SchemaInfo.
 func fixLabelNames(prov *tfbridge.ProviderInfo) {
 	// Recursively applies the label fix
-	var apply func(string, shim.Schema, *tfbridge.SchemaInfo)
-	apply = func(key string, m shim.Schema, info *tfbridge.SchemaInfo) {
+	var apply func(string, shim.Schema, *tfbridge.SchemaInfo, *[]resource.PropertyPath, resource.PropertyPath)
+
```
iwahbe, line 21: Style nit: I prefer to not have a line separation. This makes it clearer to me that we are declaring and then assigning to the same var.
```suggestion
	var apply func(string, shim.Schema, *tfbridge.SchemaInfo, *[]resource.PropertyPath, resource.PropertyPath)
```

Feel free to ignore me here if it's clearer for you.

provider/labels.go
```diff
@@ -78,3 +99,23 @@ func fixLabelNames(prov *tfbridge.ProviderInfo) {
 	}
 
 }
+
+// pulumiLabels represents a field added upstream via a custom diff, as terraform_labels.
+// There is a bug: f this field is nil in the Pulumi state, it will never be added to the state,
+// creating a permanent diff.
+// See also: https://github.com/pulumi/pulumi-gcp/issues/1314
+func ensureLabelPathsExist(paths []resource.PropertyPath) tfbridge.PropertyTransform {
+
```
iwahbe, line 108: We don't need a newline here.
```suggestion
```

provider/labels.go
```diff
@@ -78,3 +99,23 @@ func fixLabelNames(prov *tfbridge.ProviderInfo) {
...
+		for _, path := range paths {
+			// If a `pulumiLabels` field is not set at the expected path, we will set it to an empty property Map,
+			// so Diff can detect it.
+			if _, found := path.Get(obj); !found {
+				path.Set(obj, resource.NewObjectProperty(resource.PropertyMap{}))
+			}
+		}
+		return obj.ObjectValue(), nil
+	}
+
```
iwahbe, line 120: Or here
```suggestion
```

provider/labels.go
```diff
@@ -78,3 +99,23 @@ func fixLabelNames(prov *tfbridge.ProviderInfo) {
...
+// See also: https://github.com/pulumi/pulumi-gcp/issues/1314
+func ensureLabelPathsExist(paths []resource.PropertyPath) tfbridge.PropertyTransform {
+
+	return func(ctx context.Context, prop resource.PropertyMap) (resource.PropertyMap, error) {
+		obj := resource.NewObjectProperty(prop)
+		for _, path := range paths {
+			// If a `pulumiLabels` field is not set at the expected path, we will set it to an empty property Map,
+			// so Diff can detect it.
+			if _, found := path.Get(obj); !found {
+				path.Set(obj, resource.NewObjectProperty(resource.PropertyMap{}))
```
iwahbe, line 115: I wasn't sure if `Get` and `Set` would work with globs. They don't. This is fine, since there are no `pulumiLabels` through `lists` or `maps`. I'll add a suggestion for how we can assert that at `make tfgen` time.

provider/labels.go
```diff
@@ -36,40 +44,53 @@ func fixLabelNames(prov *tfbridge.ProviderInfo) {
...
 			}
 			obj.Range(func(key string, m shim.Schema) bool {
-				i, ok := info.Fields[key]
+				i, ok := info.Elem.Fields[key]
 				if !ok {
 					i = new(tfbridge.SchemaInfo)
 					info.Elem.Fields[key] = i
 				}
-				apply(key, m, i)
+				currentPath = append(currentPath, key)
```
iwahbe, line 52: We forgot a key translation.
```suggestion
				currentPath = append(currentPath,
					tfbridge.TerraformToPulumiNameV2(key, obj, info.Elem.Fields))

```

provider/labels.go
```diff
@@ -15,18 +17,24 @@ import (
...
-	apply = func(key string, m shim.Schema, info *tfbridge.SchemaInfo) {
+	var apply func(string, shim.Schema, *tfbridge.SchemaInfo, *[]resource.PropertyPath, resource.PropertyPath)
+
+	apply = func(key string, m shim.Schema, info *tfbridge.SchemaInfo,
+		paths *[]resource.PropertyPath,
+		currentPath resource.PropertyPath,
+	) {
 		if key == "terraform_labels" {
 			info.Name = "pulumiLabels"
+			// Gather all paths to pulumiLabels
```
iwahbe, line 28: As I said later in the review, `resource.PropertyPath.{Get,Set}` don't work correctly with globs, and right now they don't need to. To avoid shipping bugs to customers, lets assert here (during `make tfgen` time), that there are no globs in this path.
```suggestion
			for _, v := range currentPath {
				if v, ok := v.(string); ok && v == "*" {
					panic(fmt.Sprintf("at path %s: globs are not supported", currentPath))
				}
			}
			// Gather all paths to pulumiLabels
```

provider/labels.go
```diff
@@ -15,18 +17,24 @@ import (
...
+			*paths = append(*paths, currentPath)
```
iwahbe, line 29: This should reference the pulumi name, not the tf name.
```suggestion
			*paths = append(*paths,
				append(currentPath[:len(currentPath)-1], info.Name))
```

iwahbe (commented):
I think we missed a couple of corner cases. It's not a problem, but we should address before merging.

t0yv0:
Are there any situations where the path length is greater than 0? What are these use cases?

t0yv0:
```
#### PATH metadata.self_link.namespace.effective_annotations.effective_labels.generation.uid.annotations.labels.resource_version.terraform_labels FOR terraform_labels google_cloud_run_service
#### PATH metadata.generation.resource_version.uid.annotations.labels.effective_labels.self_link.terraform_labels FOR terraform_labels google_cloud_run_domain_mapping
#### PATH metadata.resource_version.namespace.generation.terraform_labels FOR terraform_labels google_cloud_run_service

```

t0yv0:
These are the only 3 nested cases in the entire schema. Are we even certain that those cases need patching? Cursory glance seems to indicate that the metadata/labels are for something the cloud_run_service will be managing, not for it itself.

https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/cloud_run_service

If that's right, can we stick to top-level attributes only and skip the visitor pattern/recursion schemes here? Would it make it simpler?

provider/labels.go
```diff
@@ -36,40 +44,53 @@ func fixLabelNames(prov *tfbridge.ProviderInfo) {
...
+	// - Create a collection of resource property paths that have this pulumiLabels key amd return it.
+	update := func(s shim.SchemaMap, fields *map[string]*tfbridge.SchemaInfo) []resource.PropertyPath {
+		var paths []resource.PropertyPath
 
-	update := func(s shim.SchemaMap, fields *map[string]*tfbridge.SchemaInfo) {
 		if *fields == nil {
 			*fields = make(map[string]*tfbridge.SchemaInfo, s.Len())
 		}
 		s.Range(func(key string, m shim.Schema) bool {
+			currentPath := resource.PropertyPath{key}
```
iwahbe, line 75:
```suggestion
			currentPath := resource.PropertyPath{
				tfbridge.TerraformToPulumiNameV2(key, s, fields),
			}
```

iwahbe (approved):

Issue #1316, opened 2023-11-04
[v7] Propagate secretness from `labels` to `pulumiLabels` and `effectiveLabels`


iwahbe:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

`labels` can contain secrets. If a label (or the `labels` block) is a secret, then it will be encrypted in state. We need to make sure that we see the same behavior when propagated to `effectiveLabels` and `pulumiLabels`.

iwahbe:
We have hedged to by making `pulumiLabels` and `effectiveLabels` secret: https://github.com/pulumi/pulumi-gcp/commit/ab4ff6e3db073ce78e0f12a4a0d1bead539b6bce by @guineveresaenger. 

I'm leaving this issue open for finer grained secrets when https://github.com/pulumi/pulumi-terraform-bridge/issues/1506 is implemented.

PR #1317, opened 2023-11-04
Enforce labels secretness

iwahbe:
Fixes https://github.com/pulumi/pulumi-gcp/issues/1316

This is a WIP PR. The code should be complete, but there are not any tests.

Update:

The bridge is not currently set up to accept secrets from the engine:

https://github.com/pulumi/pulumi-terraform-bridge/blob/218029989c4c3e1086b576dada10f77ccff30cf9/pkg/tfbridge/provider.go#L537-L539

https://github.com/pulumi/pulumi-terraform-bridge/blob/218029989c4c3e1086b576dada10f77ccff30cf9/pf/internal/plugin/provider_server.go#L274-L279

The code in this PR propagates secrets, but it never receives secrets to propagate. Enforcing this will require allowing the bridge to accept secrets. Doing so safely is a large project.


iwahbe:
Blocked on https://github.com/pulumi/pulumi-terraform-bridge/issues/1506.

PR #1321, opened 2023-11-06
Add Output Transformer to set aggregate labels fields Secret at runtime

guineveresaenger:
Missing a few test cases. Thinking this is the correct direction. 

Here's an example console output:

```
@guin:gcp-labels🦉 pulumi up
Previewing update (dev)

View in Browser (Ctrl+O): https://app.pulumi.com/guinevere-pulumi-corp/dev/dev/previews/5b1266d0-aaf1-4aa3-9aa3-faadf930618e

     Type                     Name          Plan       Info
 +   pulumi:pulumi:Stack      dev-dev       create     2 warnings
 +   ├─ pulumi:providers:gcp  gcp-provider  create     
 +   └─ gcp:storage:Bucket    my-bucket     create     

Diagnostics:
  pulumi:pulumi:Stack (dev-dev):
    warning: using pulumi-resource-gcp from $PATH at /Users/guin/go/src/github.com/pulumi/pulumi-gcp/bin/pulumi-resource-gcp
    warning: using pulumi-resource-gcp from $PATH at /Users/guin/go/src/github.com/pulumi/pulumi-gcp/bin/pulumi-resource-gcp

Outputs:
    bucketName       : output<string>
    pulumiAllLabels  : [secret]
    resourceSetLabels: [secret]

Resources:
    + 3 to create

Do you want to perform this update? yes
Updating (dev)

View in Browser (Ctrl+O): https://app.pulumi.com/guinevere-pulumi-corp/dev/dev/updates/75

     Type                     Name          Status              Info
 +   pulumi:pulumi:Stack      dev-dev       created (4s)        2 warnings
 +   ├─ pulumi:providers:gcp  gcp-provider  created (0.08s)     
 +   └─ gcp:storage:Bucket    my-bucket     created (2s)        

Diagnostics:
  pulumi:pulumi:Stack (dev-dev):
    warning: using pulumi-resource-gcp from $PATH at /Users/guin/go/src/github.com/pulumi/pulumi-gcp/bin/pulumi-resource-gcp
    warning: using pulumi-resource-gcp from $PATH at /Users/guin/go/src/github.com/pulumi/pulumi-gcp/bin/pulumi-resource-gcp

Outputs:
    bucketName       : "gs://my-bucket-e7310c1"
    pulumiAllLabels  : [secret]
    resourceSetLabels: [secret]

Resources:
    + 3 created

Duration: 5s
```

### DIFF
--- a/provider/labels.go
+++ b/provider/labels.go
@@ -36,6 +36,14 @@ func fixLabelNames(prov *tfbridge.ProviderInfo) {
 			currentPath = currentPath[:len(currentPath)-1]
 			currentPath = append(currentPath, info.Name)
 			*paths = append(*paths, currentPath)
+
+			// This field represents an aggregate of resource-level `labels` and provider-level `defaultLabels` fields.
+			// To avoid leaking secrets set via Input on those fields, we mark this field as unconditionally Secret.
+			info.Secret = tfbridge.True()
+		}
+		// Also set `effectiveLabels` to Secret.
+		if key == "effective_labels" {
+			info.Secret = tfbridge.True()
 		}
 
 		if m == nil {
@@ -100,7 +108,6 @@ func fixLabelNames(prov *tfbridge.ProviderInfo) {
 		// Rename pulumiLabel fields and obtain paths
 		labelPaths := update(rMap.Get(token).Schema(), &info.Fields)
 		if len(labelPaths) > 0 {
-			info.TransformOutputs = setLabelsFieldsSecret
 			info.TransformFromState = ensureLabelPathsExist(labelPaths)
 		}
 	}
@@ -128,13 +135,4 @@ func ensureLabelPathsExist(paths []resource.PropertyPath) tfbridge.PropertyTrans
 		}
 		return obj.ObjectValue(), nil
 	}
-}
-
-// If a resource has a `labels` field, the upstream provider will add the new aggregate label fields, `effectiveLabels`
-// and `terraformLabels`. To protect against accidental leaking of secrets via aggregation from resource-level `labels`
-// and provider-level `defaultLabels` fields, at runtime, we set these Outputs to Secret, if present.
-func setLabelsFieldsSecret(ctx context.Context, prop resource.PropertyMap) (resource.PropertyMap, error) {
-	prop["pulumiLabels"] = resource.MakeSecret(prop["pulumiLabels"])
-	prop["effectiveLabels"] = resource.MakeSecret(prop["effectiveLabels"])
-	return prop, nil
 }


t0yv0 (changes requested):
I'd like to offer a schema based approach instead

AWS is converging on:

```
 fields["tags_all"].Secret = &yes 
```

e.g. in https://github.com/pulumi/pulumi-aws/pull/2944/files#diff-34c57e622183cb0d8dd0d3f9eaa0861b3340120e9b2ad811bac7ac7be4cea4b1R7188

Then the bridge does the necessary marking.

AWS started off with a similar approach through output transformers, but the secret leak tests detected that the secret still can escape via Pulumi inputs which are not subject to the transformers.. 

This may or may not apply to the GCP situation, but testing it out takes time, and anyway relying on the bridge to treat a field as secret is a tad simpler as far as this repo is concerned.

provider/labels.go
```diff
@@ -128,3 +129,12 @@ func ensureLabelPathsExist(paths []resource.PropertyPath) tfbridge.PropertyTrans
 		return obj.ObjectValue(), nil
 	}
 }
+
+// If a resource has a `labels` field, the upstream provider will add the new aggregate label fields, `effectiveLabels`
+// and `terraformLabels`. To protect against accidental leaking of secrets via aggregation from resource-level `labels`
+// and provider-level `defaultLabels` fields, at runtime, we set these Outputs to Secret, if present.
+func setLabelsFieldsSecret(ctx context.Context, prop resource.PropertyMap) (resource.PropertyMap, error) {
+	prop["pulumiLabels"] = resource.MakeSecret(prop["pulumiLabels"])
+	prop["effectiveLabels"] = resource.MakeSecret(prop["effectiveLabels"])
```
iwahbe, line 138: This doesn't apply to nested secrets. I agree with @t0yv0 that we should, do this in the schema, not at runtime.

Is there a reason why something like this doesn't work:

```patch
 provider/labels.go | 3 +++
 1 file changed, 3 insertions(+)

diff --git a/provider/labels.go b/provider/labels.go
index 7fff7348c..12e84afa1 100644
--- a/provider/labels.go
+++ b/provider/labels.go
@@ -26,6 +26,7 @@ func fixLabelNames(prov *tfbridge.ProviderInfo) {
 	) {
 		if key == "terraform_labels" {
 			info.Name = "pulumiLabels"
+			info.Secret = tfbridge.True()
 			// Gather all paths to pulumiLabels
 			for _, v := range currentPath {
 				if v, ok := v.(string); ok && v == "*" {
@@ -36,6 +37,8 @@ func fixLabelNames(prov *tfbridge.ProviderInfo) {
 			currentPath = currentPath[:len(currentPath)-1]
 			currentPath = append(currentPath, info.Name)
 			*paths = append(*paths, currentPath)
+		} else if key == "effective_labels" {
+			info.Secret = tfbridge.True()
 		}
 
 		if m == nil {
```

guineveresaenger: Yeah - I was thinking if we are looking to improve Secrets handling in the future, then perhaps it doesn't entirely make sense to fully schematize the Secrets here, as a fix would mean a schema change. But at that point it's six of one, half a dozen of the other.

iwahbe (approved):
Let's make sure we have a test that `pulumiLabels` values don't show up in the state file. We might already, I'm not sure.

Other then that, LGTM

t0yv0 (approved):

guineveresaenger:
Closing via #1238

PR #1325, opened 2023-11-08
fixed test name to not clash

dixler:
Part of https://github.com/pulumi/pulumi/issues/14323

Changes the name to `TestAccWebserverPy`.

When running `make test` the test case name `TestAccWebserver` clashes with another test.


guineveresaenger (approved):
Can we add a brief description of why this change is needed, for context?

Otherwise LGTM.

PR #1335, opened 2023-11-14
Cleanup version prefix

t0yv0:
We set VERSION_PREFIX temporarily during the v7 release to make sure pulumictl versions code as 7.x.x; it is no longer necessary now that pulumictl will be finding a 7.x. tag in Git history.


thomas11 (approved):
ah, good old `VERSION_PREFIX`, confuser of developers

Issue #1353, opened 2023-11-15
Java build failure during SDK release job


t0yv0:
### What happened?

```/home/runner/work/pulumi-gcp/pulumi-gcp/sdk/java/src/main/java/com/pulumi/gcp/compute/inputs/GlobalForwardingRuleState.java:1169: error: unknown tag: pattern:
         *   @pattern: \d+(?:-\d+)?
             ^
23 errors

```

It appears that the release workflow https://github.com/pulumi/pulumi-gcp/actions/runs/6881931301/job/18720430202#step:3:6581 did not build (or publish) the Java SDK properly.

### Example

N/A

### Output of `pulumi about`

N/A

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

guineveresaenger:
A rerun resulted in the same error: https://github.com/pulumi/pulumi-gcp/actions/runs/6881931301/job/18728079843#step:3:2830

t0yv0:
Ah I see I guess we're expecting 7.1.x patch to relese but we forfeit having 7.1.0 in https://central.sonatype.com/artifact/com.pulumi/gcp/versions .. slightly unfortunate but does indeed resolve the problem. Thank you.

PR #1357, opened 2023-11-16
Patch documentation for Compute Forwarding Rule and Compute Global Forwarding Rule

guineveresaenger:
Upstream introduced what looks to be a spurious string `@pattern: \d+(?:-\d+)?` to the description for ComputeForwardingRule and ComputeGlobalForwardingRule. We need to patch this because Javadoc interprets this string as a Javadoc tag, but such a tag does not exist. This resulted in a [failure to publish the Java SDK](https://github.com/pulumi/pulumi-gcp/actions/runs/6881931301/job/18728079843) on GCP v7.1.0. 

Note that this patch _only_ addresses the documentation files. It seems likely that the upstream markdown is generated from the schema description, but as we on our end directly read from the markdown for the docs, this presents the minimum necessary change.

After this is merged, we can cut a patch version, which should result in successful publication of the Java SDK as version 7.1.1.

Follow-up issue here: https://github.com/pulumi/pulumi-java/issues/1271

Fixes https://github.com/pulumi/pulumi-gcp/issues/1353

- Add patch to remove a docstring that Java docstring does not recognize
- make schema and sdks

### DIFF
--- a/patches/0001-Allow-disabling-the-partner-name.patch
+++ b/patches/0001-Allow-disabling-the-partner-name.patch
@@ -1,7 +1,7 @@
 From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
 From: Daniel Bradley <daniel@pulumi.com>
 Date: Tue, 7 Mar 2023 11:09:59 +0000
-Subject: [PATCH 1/9] Allow disabling the partner name
+Subject: [PATCH 01/10] Allow disabling the partner name
 
 Add options to set or disable partner name.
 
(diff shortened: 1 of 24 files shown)


iwahbe (approved):
LGTM

Could we open an associated issue in pu/pu? In theory a valid schema should result in a publishable package.

guineveresaenger:
I did open https://github.com/pulumi/pulumi-java/issues/1271 as it's pretty java specific. Will edit the PR description.

PR #1364, opened 2023-11-20
Extended documentation for engineSplitTraffic

dulakm:
I believe the docs for `migrateTraffic` are not easily accessible. I had to look into Terraform provider code to understand that `patch` is being used. The docs should be more explainable now.

### DIFF
--- a/sdk/nodejs/appengine/engineSplitTraffic.ts
+++ b/sdk/nodejs/appengine/engineSplitTraffic.ts
@@ -12,6 +12,7 @@ import * as utilities from "../utilities";
  * To get more information about ServiceSplitTraffic, see:
  *
  * * [API documentation](https://cloud.google.com/appengine/docs/admin-api/reference/rest/v1/apps.services)
+ * * [API documentation for migrateTraffic](https://cloud.google.com/appengine/docs/admin-api/reference/rest/v1/apps.services/patch#query-parameters)
  *
  * ## Example Usage
  * ### App Engine Service Split Traffic


iwahbe:
Hey @dulakm. Thanks for the fix. Unfortunately, the bot is correct. Our SDKs are generated, and the documentation here is scraped from upstream (which is also generated).

For this provider especially, there isn't a clear path to add this to the docs.

dulakm:
Hello @iwahbe, thank you for your response. Do I understand correctly that the upstream is https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/app_engine_service_split_traffic and I should try to change the docs there?

Or are you already telling me that even that docs are auto-generated and I'll fail? 😅

iwahbe:
> Hello @iwahbe, thank you for your response. Do I understand correctly that the upstream is https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/app_engine_service_split_traffic and I should try to change the docs there?

More or less. We use `google-beta`, so it would be https://registry.terraform.io/providers/hashicorp/google-beta/latest/docs/resources/app_engine_service_split_traffic. 

> Or are you already telling me that even that docs are auto-generated and I'll fail? 😅

That is my understanding. From the [upstream provider](https://github.com/hashicorp/terraform-provider-google-beta#developing-the-provider):

> This repository is generated by [magic-modules](https://github.com/GoogleCloudPlatform/magic-modules). If you wish to work on the provider, you'll need to make changes in [magic-modules](https://github.com/GoogleCloudPlatform/magic-modules). Any changes made directly to this repository will likely be overwritten.

I believe that you will need to open a PR to https://github.com/GoogleCloudPlatform/magic-modules.

dulakm:
Understand. Thank you. Closing this PR. 🙂

PR #1383, opened 2023-11-30
Remove notes from upstream docs about storing secrets in plain text

mikhailshilkov:
Overhall of https://github.com/pulumi/pulumi-gcp/pull/1376

>  All arguments including the following potentially sensitive values will be stored in the raw state as plain text

While this is true in terraform, that does not apply to Pulumi. I added a function to edit docs that uses two regular expressions to find plain-text warnings and replace them with empty strings.

Fix https://github.com/pulumi/pulumi-gcp/issues/1195

### DIFF
--- a/provider/doc_edits.go
+++ b/provider/doc_edits.go
@@ -7,9 +7,10 @@ import (
 )
 
 func editRules(defaults []tfbridge.DocsEdit) []tfbridge.DocsEdit {
-	var result = append(defaults, fixupEffectiveLabels)
-	result = append(result, removeSecretsInPlainTextNote)
-	return result
+	return append(defaults,
+		fixupEffectiveLabels,
+		removeSecretsInPlainTextNote,
+	)
 }
 
 const effectiveLabels = `including the labels configured through Terraform`
@@ -33,7 +34,7 @@ var removeSecretsInPlainTextNote = tfbridge.DocsEdit{
 	Path: "*",
 	Edit: func(path string, content []byte) ([]byte, error) {
 		for _, r := range secretsInPlainTextNoteRegexps {
-			content = r.ReplaceAllLiteral(content, []byte(""))
+			content = r.ReplaceAllLiteral(content, nil)
 		}
 		return content, nil
 	},
--- a/provider/doc_edits.go
+++ b/provider/doc_edits.go
@@ -7,7 +7,9 @@ import (
 )
 
 func editRules(defaults []tfbridge.DocsEdit) []tfbridge.DocsEdit {
-	return append(defaults, fixupEffectiveLabels)
+	var result = append(defaults, fixupEffectiveLabels)
+	result = append(result, removeSecretsInPlainTextNote)
+	return result
 }
 
 const effectiveLabels = `including the labels configured through Terraform`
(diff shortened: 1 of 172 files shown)


provider/doc_edits.go
```diff
@@ -21,3 +23,18 @@ var fixupEffectiveLabels = tfbridge.DocsEdit{
...
+var secretsInPlainTextNoteRegexps = []*regexp.Regexp{
+	regexp.MustCompile(`~?> \*\*Warning:\*\* All arguments including the following potentially sensitive\nvalues will be stored in the raw state as plain text: .*\nRead more about sensitive data in state\.`),
+	regexp.MustCompile(`(?s)~?> \*\*((Warning)|(Note)):\*\* All arguments including .*as plain-text(\.)?( Read more about sensitive data in state\.)?`),
+}
+
+var removeSecretsInPlainTextNote = tfbridge.DocsEdit{
+	Path: "*",
+	Edit: func(path string, content []byte) ([]byte, error) {
+		for _, r := range secretsInPlainTextNoteRegexps {
+			content = r.ReplaceAllLiteral(content, []byte(""))
```
iwahbe, line 36: Optional nit: `r.ReplaceAllLiteral` accepts a slice and operates based on length, so this is equivalent.
```suggestion
			content = r.ReplaceAllLiteral(content, nil)
```

provider/doc_edits_test.go
```diff
@@ -29,3 +29,44 @@ func TestEffectiveLabels(t *testing.T) {
 		})
 	}
 }
+
+func TestSecretsInPlainTextNote(t *testing.T) {
```
iwahbe, line 33:
```suggestion
func TestSecretsInPlainTextNote(t *testing.T) {
	t.Parallel()
```

provider/doc_edits.go
```diff
@@ -7,7 +7,9 @@ import (
 )
 
 func editRules(defaults []tfbridge.DocsEdit) []tfbridge.DocsEdit {
-	return append(defaults, fixupEffectiveLabels)
+	var result = append(defaults, fixupEffectiveLabels)
+	result = append(result, removeSecretsInPlainTextNote)
+	return result
```
iwahbe, line 12: `append` takes `(src []T, elems ...T)`.
```suggestion
	return append(defaults, 
		fixupEffectiveLabels,
		removeSecretsInPlainTextNote,
	)
```

iwahbe (approved):
LGTM

provider/doc_edits_test.go
mikhailshilkov: Isn't `t.Parallel()` in the loop below enough?

iwahbe: They do different things:
1. `t.Parallel()` in the top level allows `TestSecretsInPlainTextNote` to run in the top level parallel pool with other parallel tests.
2. `t.Parallel()` nested allows those subtests to join the parallel pool for their supertest.

Skipping the top level parallel has each subtest of `TestSecretsInPlainTextNote` running concurrently while all other tests block.

I wasn't 100% sure, so I investigated [here](https://go.dev/play/p/LM5Og6ch9We).

PR #1472, opened 2023-12-20
Update to upstream v5.10.0

t0yv0:
Fixes https://github.com/pulumi/pulumi-gcp/issues/1455

This is a regular upstream update but has some hiccups preventing upgrade-provider automation from building this PR without intervention.

1. 0006 docs patch had a conflict, which I've resolved manually here

2. There is a new module "workbench" that needed manual configuration in resources.go

With these changes the provider builds for me in a straightforward way. Waiting for the CI check results.


iwahbe:
Looks like some data sources are missing docs:

```
error: failed to gather package metadata: problem gathering resources: 3 errors occurred:
	* could not find docs for resource 'google_workbench_instance_iam_binding'. Override the Docs property in the resource mapping. See type tfbridge.DocInfo for details.
	* could not find docs for resource 'google_workbench_instance_iam_member'. Override the Docs property in the resource mapping. See type tfbridge.DocInfo for details.
	* could not find docs for resource 'google_workbench_instance_iam_policy'. Override the Docs property in the resource mapping. See type tfbridge.DocInfo for details.
```

The current best practice is to manually confirm the docs are missing upstream, then set `DocInfo{AllowMissing: true}` on the effected resources.

t0yv0:
It's interesting that `make tfgen` worked locally, spot the differences ..

iwahbe:
> It's interesting that `make tfgen` worked locally, spot the differences ..

`upgrade-provider` sets `PULUMI_MISSING_DOCS_ERROR=true PULUMI_EXTRA_MAPPING_ERROR=true`, as does CI.

t0yv0:
Great pointer. This feels odd, so I filed  https://github.com/pulumi/ci-mgmt/issues/754 

But I can proceed with this tyvm.

t0yv0:
Also seeing

```
warning: parseImports "gcp:workstations/workstation:Workstation" should not render the string 'terraform' in its emitted markdown.
```

We should probably fix these..

t0yv0:
IN fact https://github.com/pulumi/pulumi-gcp/issues/1473 there's a lot of these docs issues, I'm going to treat as a separate ticket. Looks like it would not be good to rely on patches for these like the one I just edited but instead have some form of autocorrect (at this scale).

iwahbe (approved):

PR #1555, opened 2024-01-24
Clean up Patches containing `Beta` links

### LINKED ISSUE #1540: Patch cleanup: Docs
In https://github.com/pulumi/pulumi-terraform-bridge/pull/1629, we have improved the stripping of Terraform links.

This means that any resource field descriptions of that looked somewhat like this:
`(Optional, [Beta](https://terraform.io/docs/providers/google/guides/provider_versions.html))`

were omitted from our documentation, [unless manually patched](https://github.com/pulumi/pulumi-gcp/blob/master/patches/0006-docs-patching.patch#L570C4-L570C96).

Patching up these links is a no-win, since their need is impossible to detect (tfgen just drops the whole description) and the maintenance via patch is manual and slow.

With the bridge changes, we will re-gain some of these fields; however they may contain confusing `Beta` references, which we do not want.

In order to reduce the maintenance burden and increase docs accuracy, I suggest the following:

1. Remove all patches containing the `[Beta](https://terraform.io/docs/providers/google/guides/provider_versions.html)` fixup, relying on the above bridge changes for link stripping
2. To remove the non-Pulumi-applicable `Optional, Beta` terms at the beginning of descriptions, create a special `DocsRule` for GCP only. This rule should match any of the following, with or without parens:
   `Optional, Beta` 
   `Beta `
   `Optional, Beta, Deprecated`
Note that the specific matching details here depend on the surrounding `EditRule`s and may present slightly differently.

guineveresaenger:
This pull request cleans up manual patch entries that remove links of the format `[Beta](https://terraform.io/docs/providers/google/guides/provider_versions.html)` in favor of [new bridge functionality](https://github.com/pulumi/pulumi-terraform-bridge/pull/1629) that can strip those links.

To clean up the danging `Beta` references unique to this provider, a new `docsEditRule` and corresponding tests were added.

~This PR is pinned to an unreleased version of the bridge, so no SDKs were generated and this PR is a draft. Nonetheless, the schema is checked in here to show overall improvements compared with current default branch.~ PR is now rebased against current default branch.

Fixes #1540.


provider/doc_edits.go
```diff
@@ -39,3 +40,31 @@ var removeSecretsInPlainTextNote = tfbridge.DocsEdit{
...
+// Some descriptions had a link to google-beta removed as part of link scrubbing by the bridge.
+// This leaves us with the following patterns unique to this provider to remove from the "Description" fields.
+// We have to be somewhat verbose here to avoid removing legitimate use of the term "Beta", e.g "Kubernetes Beta API".
+// "(Optional, Beta, Deprecated)
+// "(Optional, Beta)
+// "(Beta, Deprecated)
+// "(Beta only)
+// "(Optional) Beta
+var betaRegexps = []*regexp.Regexp{
+
```
iwahbe, line 53: Nit: Extra space.
```suggestion
```

iwahbe (approved):
This looks like a big improvement in our docs. LGTM


I'm seeing lots of syntactic cleanup like:

```patch
-                    "description": ")\nProperties to set on all instances in the group. After setting\nallInstancesConfig on the group, you must update the group's instances to\napply the configuration.\n"
+                    "description": "Properties to set on all instances in the group. After setting\nallInstancesConfig on the group, you must update the group's instances to\napply the configuration.\n"
```

I'm seeing new docs:

```patch
                 "keepCount": {
-                    "type": "integer"
+                    "type": "integer",
+                    "description": "Minimum number of versions to keep.\n"
                 },
```

And of course, we are now correctly removing the "Beta" from our docs:

```patch
-                    "description": "(Beta only) Details of placement information for an instance.\n",
+                    "description": "Details of placement information for an instance.\n",
```

iwahbe:
I'm assuming that the plan is to wait until the next bridge release, then finish the PR and merge.

PR #1579, opened 2024-01-26
Upgrade providertest

t0yv0:
Switching to the latest recommended API version for providertest. Borrowing some recordings (cluster) from @VenelinMartinov PR. Four tests are still skipped pending https://github.com/pulumi/pulumi-gcp/pull/1603 and one test is newly skipped because it seems to indicate an actual failure.


provider/test-programs/network/Pulumi.yaml
```diff
@@ -5,22 +5,30 @@ resources:
   exampleNetwork:
     type: gcp:compute/network:Network
     properties:
-      name: pulumi-network
+      name: mynetwork2
+        # Randomizing the name fails with "  warning: Error downloading plugin: unknown version for plugin str"
```
t0yv0, line 9: This was really unexpected and annoying.

provider/test-programs/network/Pulumi.yaml
```diff
@@ -5,22 +5,30 @@ resources:
...
+        #     new: ""
+        #     string: network-${pulumi.stack}
+        #   return: result
       autoCreateSubnetworks: false  # We'll create our own subnetwork
 
   exampleSubnetwork:
     type: gcp:compute/subnetwork:Subnetwork
     properties:
-      name: pulumi-subnetwork
+      name: pulumi-subnetwork2
```
t0yv0, line 22: The names are not randomized means that if the previous tests' run did not clean up we get name collisions.

provider/testdata/recorded/TestProviderUpgrade/compute-firewall/6.67.0/stack.json
```diff
@@ -0,0 +1,173 @@
+{
```
t0yv0, line 1: The new framework expects stack.json instead of state.json ...

danielrbradley: Bah, my bad. We could rename that back for easier migration if it was a pain?

danielrbradley (approved):
This looks good! Would be good to know if there were any rough edges.

Currently, the provider version switching is only implemented by switching the provider binary but not the actual SDK version. Just checking you're happy with using it without that for now.

provider/testdata/recorded/TestProviderUpgrade/compute-firewall/6.67.0/stack.json
t0yv0: We don't have a big enough deployment base of the old one to bother, I think I can just rename.

t0yv0:
Not much in terms of friction log - some more logging would be good I guess, I don't understand why for example running the upgrade test preview even when the recordings are available ends up installing the baseline version of provider in plugins.. Still a little bit of that magical feeling. But certainly no worse than the predecessor code so it's improving. 

One issue Venelin ran into is the need to test an upgrade combined with a code edit, this seems a bit unnatural right now.

PR #1581, opened 2024-01-26
Add lowercase autonaming for subnet and router

### LINKED ISSUE #1576: AutoNaming on gcpNetwork does not pass validators
### What happened?

The default name selected for the Network resource cannot provision.

```
            gcp:compute:Network exampleNetwork  error: gcp:compute/network:Network resource 'exampleNetwork' has a problem: "name" ("exampleNetwork-12fde8d") doesn't match regexp "^(?:[a-z](?:[-a-z0-9]{0,61}[a-z0-9])?)$". Examine values at 'exampleNetwork.name'.
         +  pulumi:pulumi:Stack network-test created (1s) 
            gcp:compute:Network exampleNetwork **failed** 1 error
        Diagnostics:
          gcp:compute:Network (exampleNetwork):
            error: gcp:compute/network:Network resource 'exampleNetwork' has a problem: "name" ("exampleNetwork-12fde8d") doesn't match regexp "^(?:[a-z](?:[-a-z0-9]{0,61}[a-z0-9])?)$". Examine values at 'exampleNetwork.name'.
        
```

### Example

```
description: A minimal Google Cloud Pulumi YAML program
resources:
  exampleNetwork:
    type: gcp:compute/network:Network
    properties:
      autoCreateSubnetworks: false  # We'll create our own subnetwork

  exampleSubnetwork:
    type: gcp:compute/subnetwork:Subnetwork
    properties:
      name: pulumi-subnetwork
      network: ${exampleNetwork.id}
      ipCidrRange: "10.2.0.0/16"
      region: us-central1

  exampleRouter:
    type: gcp:compute/router:Router
    properties:
      name: pulumi-router
      network: ${exampleNetwork.id}
      region: us-central1
      bgp:
        asn: 64514
```

### Output of `pulumi about`

gcp         resource   7.7.0    208 MB  2 hours ago   2 hours ago

### Additional context

N/A

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

iwahbe:
Fixes #1576 

Following in the footsteps of https://github.com/pulumi/pulumi-gcp/commit/c7b0d6626ff3341e8ad27f039b8d31af396254b8, this PR applies lowercase auto-naming to 2 new resources: `gcp:compute:Subnetwork` and `gcp:compute:Router`.


provider/resources.go
```diff
@@ -323,6 +323,12 @@ func lowercaseAutoName() *tfbridge.SchemaInfo {
 	})
 }
 
+func lowercaseAutoNameFields() map[string]*tfbridge.SchemaInfo {
```
guineveresaenger, line 326: for readability purposes, I don't love this abstraction since it really only applies to the `name` field.
Currently, it's immediately obvious at a glance per resource that we're doing something about the `name` field specifically.

provider/resources.go
```diff
@@ -1241,13 +1239,11 @@ func Provider() tfbridge.ProviderInfo {
 			"google_compute_region_ssl_policy":                          {Tok: gcpResource(gcpCompute, "RegionSslPolicy")},
 			"google_compute_resource_policy":                            {Tok: gcpResource(gcpCompute, "ResourcePolicy")},
 			"google_compute_route":                                      {Tok: gcpResource(gcpCompute, "Route")},
-			"google_compute_router":                                     {Tok: gcpResource(gcpCompute, "Router")},
+			"google_compute_router":                                     {Fields: lowercaseAutoNameFields()},
```
guineveresaenger, line 1242: Did you leave out the `Tok` field on purpose? It looks like auto token mapping takes care of this - but then should we be consistent and remove the explicit token mapping from `Network` and the other refactored resources as well?

guineveresaenger (approved):
Looks good in functionality; a couple nits.

provider/resources.go
iwahbe: > Did you leave out the Tok field on purpose?

Yes

> It looks like auto token mapping takes care of this - but then should we be consistent and remove the explicit token mapping from `Network` and the other refactored resources as well?

Sure.

provider/resources.go
iwahbe: I don't want to duplicate out the entire `map[string]*tfbridge.SchemaInfo{ "name": value }`. It's super long and a great way to introduce a typo. I refactored `lowercaseAutoNameFields()` to `nameField(lowercaseAutoName())`, which is more explicit.

Issue #1643, opened 2024-02-07
TestAccAppServiceCs and TestAccServerless started failing


t0yv0:
### What happened?

Looks like TestAccAppServiceCs and TestAccServerless started failing the check returning 404 instead of the expected value. The failure started happening outside of any particular PR, so this either indicates flakiness or some change to the underlying system. 

### Example

See recent master builds.

### Output of `pulumi about`

N/A

### Additional context

Going to have a quick look here to reproduce locally, in case this is an actual regression. If does not look serious will temporarily skip the tests. 

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

t0yv0:
https://github.com/pulumi/pulumi-gcp/pull/1639

PR #1645, opened 2024-02-07
Skip failing tests

t0yv0:
Temporarily skipping https://github.com/pulumi/pulumi-gcp/issues/1643 until I can get credentials sorted to reproduce and debug these failures.


t0yv0:
Instead of skipping, we changed retry to accept 404. Should be better now.

PR #1713, opened 2024-02-27
Update bridge to 3.76.0

t0yv0:
This is a routine update, done by hand as the automated one has been failing.


VenelinMartinov (approved):
Thanks for sorting this out

PR #1824, opened 2024-03-13
Upgrade terraform provider google beta to v5.20.0

t0yv0:
Fixes https://github.com/pulumi/pulumi-gcp/issues/1816

The automated upgrade-provider job failed because several new resources did not have docs auto-discovered. These are now workaround by listing these resources as OK to proceed without docs.

```
error: failed to gather package metadata: problem gathering resources: 3 errors occurred:
	* could not find docs for resource 'google_clouddeploy_custom_target_type_iam_binding'. Override the Docs property in the resource mapping. See type tfbridge.DocInfo for details.
	* could not find docs for resource 'google_clouddeploy_custom_target_type_iam_member'. Override the Docs property in the resource mapping. See type tfbridge.DocInfo for details.
	* could not find docs for resource 'google_clouddeploy_custom_target_type_iam_policy'. Override the Docs property in the resource mapping. See type tfbridge.DocInfo for details.
	
```

Diving a bit deeper there is a strange pattern where three resources are documented in a single markdown file (and shown on a single page in TF registry), this one:

./upstream/website/docs/r/clouddeploy_custom_target_type_iam.html.markdown

The file itself does not correspond to any one of these resources but to all three together. We don't yet support this pattern but we could experiment with broadcasting this file as the "source" for every three resources and see how that projects into the docs.


iwahbe (approved):
LGTM

PR #2154, opened 2024-07-03
Automatically detect iam resources on tfgen

guineveresaenger:
Split out from #2153.

Fixes https://github.com/pulumi/pulumi-gcp/issues/2046..

Additionally this seems to fix a fair bit of missing or incorrect documentation, where the manual mapping had an incorrect file target.


mjeffryes (approved):
Nice! Checked all the ts docs changes and they all look like improvements to me!

PR #2490, opened 2024-10-01
Do not depend on specific sdk

t0yv0:
Re: https://github.com/pulumi/ci-mgmt/issues/1091

### DIFF
--- a/sdk/dotnet/Pulumi.Gcp.csproj
+++ b/sdk/dotnet/Pulumi.Gcp.csproj
@@ -45,7 +45,7 @@
   </ItemGroup>
 
   <ItemGroup>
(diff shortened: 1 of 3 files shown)


mjeffryes (approved):

sdk/dotnet/Pulumi.Gcp.csproj
```diff
@@ -45,7 +45,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Pulumi" Version="3.*" />
+    <PackageReference Include="Pulumi" Version="[3.66.1.0,4)" />
```
danielrbradley, line 48: I don't think this is correct in .NET as using a specific range will make nuget resolve to the **lowest** possible version.

PR #2587, opened 2024-11-06
Adding in autonaming for serviceAccount/Role, fixes #91

### LINKED ISSUE #91: Add randomness to service account and IAM IDs
There are several properties in the GCP provider that must be globally unique but aren't `name` and so don't enjoy our auto-naming suffixing. For instance, `gcp.projects.IAMCustomRole` has `roleId` and `gcp.serviceAccount.Account` has `accountId`. Although it's possible to use something like our `@pulumi/random` package to add some safe randomness to the names, it'd be nicer if it happened automatically.

rshade:
fixes #91

### DIFF
--- a/devbox.lock
+++ b/devbox.lock
@@ -1,1 +1,492 @@
-{}
+{
+  "lockfile_version": "1",
+  "packages": {
+    "curl@8": {
+      "last_modified": "2024-10-29T02:50:45Z",
+      "resolved": "github:NixOS/nixpkgs/30c9efeef01e2ad4880bff6a01a61dd99536b3c9#curl",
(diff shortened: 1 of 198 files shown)


iwahbe:
Hi @rshade. Thanks for the PR. Can you re-run SDK generation: `make tfgen build_sdks`. Auto-named properties are dropped from examples, so the examples are now out of date.

rshade:
@iwahbe any idea on that java error?

rshade:
Also from the size of the changes, I think there might be something wrong with the bridge, @iwahbe, thoughts, or did I do it wrong?

PR #2667, opened 2024-11-20
Move our registry docs into `docs/`

iwahbe:
Guides are auto-generated, so they remain in the registry.

These are direct copy + paste from the registry docs. I have not changed the prose.


guineveresaenger (approved):
LGTM - thank you!

PR #2722, opened 2024-12-10
Don't run `go test` in JSON mode

iwahbe:
Follow-up on https://github.com/pulumi/pulumi-gcp/pull/2712.


VenelinMartinov (approved):
Thanks for fixing

PR #3133, opened 2025-03-24
Draft: Workaround for broken Python dependency chain

guineveresaenger:
Do not merge as-is; draft only.


guineveresaenger:
Resolved upstream.

PR #3150, opened 2025-04-03
chore: fix .gitignore to not ignore schema.go

t0yv0:
There are files under sdk that match schema.go they probably should not be ignored.


guineveresaenger (approved):
Do we know why this was added to the gitignore?

t0yv0:
Not sure. Also schema checks are going haywire?

t0yv0:
master just failed. Disappointing.

PR #3153, opened 2025-04-08
Accommodate breaking changes in pulumi/pulumi for `property.*`

iwahbe:
This needs to go in at the same time as the version of pulumi/pulumi is bumped to v3.161.0. It reflects the breaking changes added in https://github.com/pulumi/pulumi/pull/17500.


guineveresaenger (approved):
LGTM after linter changes I think. Not sure I fully understand the upstream change.

iwahbe:
We are unable to merge this now, since due to other breaking changes in pulumi/pulumi, the bridge will need to be updated to v3.161.0 first. This PR needs to go in at the same time as the bridge update which must coincide with the pulumi/pulumi update.

t0yv0:
I think @guineveresaenger merged a version of this.

PR #3197, opened 2025-04-30
Upgrade pulumi-terraform-bridge-to v3.107.0

guineveresaenger:
This is a manual bridge dependency upgrade as there were breaking changes to providertest's handling of flags.

See also https://github.com/pulumi/pulumi-gcp/pull/3191.

Fixes https://github.com/pulumi/pulumi-gcp/issues/3194.

- **upgrade bridge version**
- **Incorporate adjustments for providertest breaking changes**
- **make schema and sdks**


provider/benchmark_test.go
```diff
@@ -23,9 +22,7 @@ import (
 // Runs a benchmark test using the provider BenchmarkConfig
 func RunBenchmark(t *testing.T, bench BenchmarkConfig) {
 	t.Run("benchmark", func(t *testing.T) {
-		if !(BenchmarkFlag.IsSet() || RecordBenchmarkFlag.IsSet()) {
-			t.Skipf("Skip benchmarking because %s", BenchmarkFlag.WhyNotSet())
-		}
+		t.Skip("Skip benchmarking because")
```
t0yv0, line 25: Am I reading this correctly that this will never run now? C

provider/labels.go
```diff
@@ -31,17 +31,16 @@ func fixEmptyLabels(_ context.Context, req shimv2.PlanStateEditRequest) (cty.Val
...
+			programLabels = labels.AsMap()
 		}
 	}
 
 	// Apply labels next, allowing labels to override defaultLabels.
 	if inputs, ok := (resource.PropertyPath{labelsPropertyName}.Get(resource.NewProperty(req.NewInputs))); ok {
 		if labels := resource.FromResourcePropertyValue(inputs); labels.IsMap() {
-			for k, v := range labels.AsMap() {
-				programLabels[k] = v
+			for k, v := range labels.AsMap().AsMap() {
```
t0yv0, line 42: This looks suspect, is there some context on what is changing for label handling? Thank you.

Zaid-Ajaj: The behavior hasn't changed but the public API of `property.Value` has. You could no longer iterate over `property.Map` directly (the result of `labels.AsMap()`) but you need to call `.AsMap()` again so that the expression becomes `map[string]property.Value` which you can iterate on

provider/benchmark_test.go
Zaid-Ajaj: `BenchmarkFlag` and `RecordBenchmarkFlag` were set to a hardcoded value where `IsSet` always returned `false` so these tests never ran (in CI)

Zaid-Ajaj: Old code looked like this:
```go
// providertest doesn't expose a real flags implementation, so just pretend for now
var (
	//("benchmark", "Run benchmarks of provider test programs and compare with recorded values")
	BenchmarkFlag = hardcodedFlag{set: false}
	//("record-benchmark", "Run benchmarks of provider test programs and record results")
	RecordBenchmarkFlag = hardcodedFlag{set: false}
)

type hardcodedFlag struct {
	set bool
}

var _ = flags.Flag(hardcodedFlag{})

func (h hardcodedFlag) IsSet() bool       { return h.set }
func (h hardcodedFlag) WhySet() string    { return "hardcoded" }
func (h hardcodedFlag) WhyNotSet() string { return "hardcoded" }
```

provider/labels.go
guineveresaenger: yeah! so @Zaid-Ajaj may have a bit more context here, but I believe it's related to the change in https://github.com/pulumi/pulumi/pull/17500.

The first `AsMap` returns a `Map` from Value but as of the above change to pu/pu, we can no longer range over a `Map`. We must retrieve the underlying map (also called `AsMap()`) in order to retrieve a map[string]Value that we can range over.

provider/labels.go
```diff
@@ -31,17 +31,16 @@ func fixEmptyLabels(_ context.Context, req shimv2.PlanStateEditRequest) (cty.Val
 
 	// Apply default labels first.
 	if pConfig := resource.FromResourcePropertyValue(resource.NewProperty(req.ProviderConfig)); pConfig.IsMap() {
-		l := pConfig.AsMap()["defaultLabels"]
-		if l.IsMap() {
-			programLabels = l.AsMap()
+		if labels, ok := pConfig.AsMap().GetOk("defaultLabels"); ok && labels.IsMap() {
```
t0yv0, line 34: Curious what is going on here.

t0yv0 (approved):

PR #3212, opened 2025-05-09
Enable RawStateDelta

t0yv0:
A bridge feature for preserving Terraform raw state is being rolled out to this provider. See also:

- https://github.com/pulumi/pulumi-terraform-bridge/issues/1667
- https://github.com/pulumi/pulumi-terraform-bridge/issues/3003

### DIFF
--- a/provider/resources.go
+++ b/provider/resources.go
@@ -2864,6 +2864,7 @@ func Provider() tfbridge.ProviderInfo {
 		},
 		EnableZeroDefaultSchemaVersion: true,
 		EnableAccurateBridgePreview:    true,
+		EnableRawStateDelta:            true,
 	}
 
 	prov.RenameResourceWithAlias("google_compute_managed_ssl_certificate", gcpResource(gcpCompute,


VenelinMartinov (approved):

t0yv0:
Investigating the issue. The diff seems to be a miss om big integer rounding.

When the state for gcp:compute:BackendService  get written the back-read check fails. 

```
-  "generated_id": 2495055709147741000,
+  "generated_id": 2495055709147741188,
```

rawStateRecoveredBytes has 2495055709147741000 instead of the expected 2495055709147741188.

t0yv0:
```
"generated_id": {
  Type:        schema.TypeInt,
  Computed:    true,
  Description: `The unique identifier for the resource. This identifier is defined by the server.`,
},
```


It got rounded when turning to PropertyValue:

```
   generatedId:{2.495055709147741e+18}
```

Which is unfortunate; possibly bridge should have turned it into a string encoding that preserves the bits instead of truncating. That said I remember a precedent of relaxing the write-through check to tolerate rounding. Looking up the details.

t0yv0:
Waiting on bridge release: https://github.com/pulumi/pulumi-terraform-bridge/issues/3093

Issue #3328, opened 2025-08-16
Errors with autogenerated roleIds containing hyphens


maelp:
### Describe what happened

I'm trying to create a role:

```ts
const readonlyRole = new gcp.organizations.IAMCustomRole(
    "gouach-public-assets-read-only",
    {
      orgId: config.require("gcpOrganizationId"),
      title: "Gouach Public Assets bucket Read-only policy",
      description: "Allow only storage.objects.get on the public-assets bucket",
      permissions: ["storage.objects.get"],
      stage: "GA",
    },
  );
```

and it generates a `roleId` containing an hyphen, although GCP doesn't allow this:

```
error: gcp:organizations/iAMCustomRole:IAMCustomRole resource 'gouach_public_assets_read_only' has a problem: "role_id" ("gouach_public_assets_read_only-d35e3d9") doesn't match regexp "^[a-zA-Z0-9_\\.]{3,64}$". Examine values at 'gouach_public_assets_read_only.roleId'.
```

in order to bypass this, I have to explicitly set `{... roleId: "my_role_id"}` in the args, but I guess the `@pulumi/gcp` plugin should not try to use hyphens in roles if that's disallowed by GCP?

### Sample program

```ts
const readonlyRole = new gcp.organizations.IAMCustomRole(
    "gouach-public-assets-read-only",
    {
      orgId: config.require("gcpOrganizationId"),
      title: "Gouach Public Assets bucket Read-only policy",
      description: "Allow only storage.objects.get on the public-assets bucket",
      permissions: ["storage.objects.get"],
      stage: "GA",
    },
  );
```

### Log output

error: gcp:organizations/iAMCustomRole:IAMCustomRole resource 'gouach_public_assets_read_only' has a problem: "role_id" ("gouach_public_assets_read_only-d35e3d9") doesn't match regexp "^[a-zA-Z0-9_\\.]{3,64}$". Examine values at 'gouach_public_assets_read_only.roleId'.

### Output of `pulumi about`

CLI          
Version      3.153.1
Go Version   go1.23.6
Go Compiler  gc

Plugins
KIND      NAME     VERSION
resource  command  1.1.0
resource  gcp      8.40.0
language  nodejs   3.153.1

Host     
OS       darwin
Version  15.5
Arch     arm64

This project is written in nodejs: executable='/Users/primet/.local/share/mise/installs/node/22.16.0/bin/node' version='v22.16.0'

Current Stack: maelp/gouach-acacia-gcp/production

[...]

PR #3331, opened 2025-08-18
Update resources.go to use underscore for role_ids

maelp:
Solves https://github.com/pulumi/pulumi-gcp/issues/3328


blampe (approved):
Thank you!

Issue #2307, opened 2024-08-08
CloudSQL instances with auto disk resizing fail to update on the first try but succeed on retrying


parth-da:
### Describe what happened

CloudSQL for Postgres instances created with `diskAutoresize: true` always fail if the size of the database has changed since the last time they were updated via pulumi with:

```
    error: 1 error occurred:
    	* updating urn:pulumi:cluster.scratchc::cluster::gcp:sql/databaseInstance:DatabaseInstance::test-pg: 1 error occurred:
    	* Error, failed to update instance settings for : googleapi: Error 400: Invalid request: The disk size cannot decrease. Current size: 31 GB, requested: 10 GB.., invalid
```

Re-running `pulumi up` however succeeds.

It looks like pulumi caches the old disk size (not sure where since I couldn't find it in the output of `pulumi stack export`) and sends it to GCP which then rejects the request. The first run updates this cache allowing subsequent attempts to succeed.

This is very similar to https://github.com/pulumi/pulumi-gcp/issues/549 except for the `ignoreChanges` part.

### Sample program

Do the same steps as in https://github.com/pulumi/pulumi-gcp/issues/549#issuecomment-810492088 (removing the `ignoreChanges` lines).

Then re-run `pulumi up` after making some small change to the script. I was trying to set a database flag.

```patch
       databaseVersion: "POSTGRES_13",
+      databaseFlags: [{"temp_file_limit", 500000}],
       settings: {
```

### Log output

```
    error: 1 error occurred:
    	* updating urn:pulumi:cluster.scratchc::cluster::gcp:sql/databaseInstance:DatabaseInstance::test-pg: 1 error occurred:
    	* Error, failed to update instance settings for : googleapi: Error 400: Invalid request: The disk size cannot decrease. Current size: 31 GB, requested: 10 GB.., invalid
```

### Affected Resource(s)

CloudSQL for Postgres instances

### Output of `pulumi about`

```
CLI
Version      3.112.0
Go Version   go1.22.1
Go Compiler  gc

Plugins
NAME    VERSION
nodejs  unknown

Host
OS       darwin
Version  14.6.1
Arch     arm64

This project is written in nodejs: executable='/nix/store/9zli090ri6wlhjla6bb51dg326ann92x-nodejs-20.12.2/bin/node' version='v20.12.2'
```

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

[...]

PR #3369, opened 2025-09-08
fix: Change the placement of the unconditional read in GCP DB Instance Update

same-id:
Currently there are some code paths where we do not read the settings before an update.

Fix the misplaced read.

See:
  https://github.com/GoogleCloudPlatform/magic-modules/pull/14952
  
CC: @AaronFriel @guineveresaenger 

Fixes: https://github.com/pulumi/pulumi-gcp/issues/2307


same-id:
Nevermind this PR it won't fix the DB issue due to the `customdiff.ForceNewIfChange("settings.0.disk_size", compute.IsDiskShrinkage),` line that runs even before the code in the Update does, a more thorough change is required here just like in https://github.com/GoogleCloudPlatform/magic-modules/pull/14952

I can patch it instead.

Closing this PR for now

PR #3370, opened 2025-09-09
Remove unnecessary link stripping from website-docs-d-tweaks patch

guineveresaenger:
So.... the bridge actually has fairly sophisticated link stripping functionality built in already.

This pull request removes redundant link stripping from the data source docs patching patch.

The schema should generate with zero diff. 

I'm happy to PR this into `master` if we prefer.


Zaid-Ajaj (approved):
Nice!

PR #3469, opened 2025-11-20
Peg cloudfunction mixin runtime to build-time NodeJS version."

### LINKED ISSUE #3460: Set the default cloudfunction NodeJS runtime version to the NodeJS build time version
We provide a cloudfunction mixin for NodeJS that gives Pulumi users the ability to write functions inline and serialize the closure. To help Pulumi users, we made the NodeJS runtime version an optional parameter. Hence, we are hardcoding an arbitrary NodeJS runtime version:

https://github.com/pulumi/pulumi-gcp/blob/2bcc590a777e1de2f6f02155c51c4e432f11734a/sdk/nodejs/cloudfunctions/zMixins.ts#L170-L176

Unfortunately, when a version of NodeJS is decommissioned because it reached it end-of-life, we need to remember to modify this value to the next LTS version. This can potentially lead to a misleading behavior on Pulumi users as we are updating the NodeJS version without our users realizing. This issue was raised [on this PR comment](https://github.com/pulumi/pulumi-gcp/pull/3456#discussion_r2522764481).

Alternatively, since the original use case was _write functions inline and serialize the closure_, we could set that runtime value based on the current NodeJS version that's executing the Pulumi program by using the `process.version` to set that value. The code will end up looking something like this:

```ts

// process.version output looks like  v20.19.5
const majorVersion = process.version.slice(1).split('.')[0];
const runtimeVersion = `nodejs${majorVersion}`;

this.function = new cloudfunctions.Function(functionName, {
   ...
  runtime: utils.ifUndefined(args.runtime, runtimeVersion),
}, parentOpts);
```

This approach not only means we need to update less frequently this code, but also gives back control to the Pulumi user: Whenever they update the NodeJS version they use to run Pulumi, they will get also that runtime version on their Lambdas (leading to less astonishment).

pose:
Fixes #3460 

This PR sets that CloudFunction runtime value based on the current NodeJS version that's executing the Pulumi program by using the `process.version` to set that value. 

This approach not only means we need to update less frequently this code, but also gives back control to the Pulumi user: Whenever they update the NodeJS version they use to run Pulumi, they will get also that runtime version on their CloudFunctions (leading to less astonishment).


sdk/nodejs/cloudfunctions/zMixins.ts
```diff
@@ -167,9 +167,13 @@ export class CallbackFunction extends pulumi.ComponentResource {
...
         const functionName = name.replace(/[^-_a-z0-9]/gi, "-");
 
+        // Extract major version from process.version (e.g., "v20.19.5" -> "20")
+        const majorVersion = process.version.slice(1).split('.')[0];
+        const defaultRuntime = `nodejs${majorVersion}`;
+
         this.function = new cloudfunctions.Function(functionName, {
             ...argsCopy,
-            runtime: utils.ifUndefined(args.runtime, "nodejs22"),
+            runtime: utils.ifUndefined(args.runtime, defaultRuntime),
```
pose: Core logic to review 👀

Graham-Pedersen (approved):
LGTM no suggested changes.

PR #3530, opened 2026-01-14
chore: bump the runner size for build_sdk

pgavlin:
This should help fix failures like
https://github.com/pulumi/pulumi-gcp/actions/runs/20981775646/job/60404174224?pr=3528.


guineveresaenger (approved):
whoa, has the ci-mgmt pr been failing for a while? 

Anyway nothing sticks out. :shipit:

PR #3826, opened 2026-06-08
fix: remove broken GKE managed ML diagnostics doc link

immanuwell:
Fixes a small docs papercut.

`gcp.container.Cluster` shipped a dead `https://docs.cloud.google.com/kubernetes-engine/docs/concepts/TODO` link and an extra leading `)` in the `managedMachineLearningDiagnosticsConfig` description. This keeps public docs clean, while the test still documents that upstream currently ships a TODO placeholder there. No runtime behavior change, just docs.

Repro:
1. On `master`, search `managedMachineLearningDiagnosticsConfig` in `provider/cmd/pulumi-resource-gcp/schema.json`.
2. Open `https://docs.cloud.google.com/kubernetes-engine/docs/concepts/TODO`.
3. It returns `404`, so the link is busted.

Checks:
- `cd provider && go test . -run "TestReplacementsDotJSON|TestReplacementDoesNotIncludeTodos"`
- `python3 -m py_compile sdk/python/pulumi_gcp/container/cluster.py`
- `cd sdk && go build ./go/gcp/container`

### DIFF
--- a/provider/doc_edits_test.go
+++ b/provider/doc_edits_test.go
@@ -43,6 +43,8 @@ func TestReplacementsDotJSON(t *testing.T) {
 	require.NoError(t, err)
 	assert.Equal(t, "Terraform should be untouched here.", string(actual))
 
+	// Upstream currently ships a placeholder TODO URL here. We should not surface
+	// that broken internal marker in generated user-facing docs.
 	actual, err = edit.Edit("container_cluster.html.markdown", []byte(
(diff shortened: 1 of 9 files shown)


guineveresaenger:
This is a fix that should be made against the upstream Google Terraform Provider.

We regret the broken link but do not believe that its removal improves our user experience in any way.

PR #3860, opened 2026-06-29
Forward universeDomain, billingProject, and userProjectOverride to Terraform

Tomahna:
`universeDomain`, `billingProject` and `userProjectOverride` are exposed in the Pulumi schema's but never reach the underlying Terraform provider. They're missing from the `Config:` map in `provider/resources.go`.

Expected behavior
----
Having these configuration knob setup in a pulumi stack should allow working in a non default google universe / non default project.
```
gcp:universeDomain: ...
gcp:billingProject: ...
gcp:userProjectOverride: ...
```

Current behavior
----
```
  error: Running program '...' failed with an unhandled exception                                                                                                                       
  Error: error serializing property "members": invocation of gcp:organizations/getProject:getProject returned an error: 1 error occurred:                                               
      * Universe domain mismatch: 'googleapis.com' does not match the universe domain '<configured-universe>' supplied directly to Terraform.
```

### DIFF
--- a/provider/resources.go
+++ b/provider/resources.go
@@ -555,6 +555,21 @@ func Provider() tfbridge.ProviderInfo {
 			"terraform_attribution_label_addition_strategy": {
 				Name: "pulumiAttributionLabelAdditionStrategy",
 			},
+			"universe_domain": {
+				Default: &tfbridge.DefaultInfo{
+					EnvVars: []string{"GOOGLE_UNIVERSE_DOMAIN"},
+				},
+			},
+			"billing_project": {
+				Default: &tfbridge.DefaultInfo{
+					EnvVars: []string{"GOOGLE_BILLING_PROJECT"},
+				},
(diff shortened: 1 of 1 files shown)


guineveresaenger:
Hi @Tomahna, thank you for your pull request and we're sorry you're running into trouble.

Explicit mapping in resources.go is not necessary to read and forward environment variables. All config env vars are read and forwarded in https://github.com/pulumi/pulumi-terraform-bridge/blob/v3.134.0/pkg/tfbridge/provider.go#L721-L740. This change would add no new behavior.

The fact that you're seeing that particular error tells me that your domain var *is* reaching Terraform. My guess is that it's a credentials mismatch - your credentials are resolving to `googleapis.com` not your configured universe domain.

If checking your credentials does not lead to success, please file a bug report with a minimal repro showing the bug and we'll be happy to take a look!

PR #3932, opened 2026-08-21
Add scripts/new_resources to resolve upstream new resources to their products

pose:
Walks a terraform-provider-google-beta release tag to the magic-modules product each new resource belongs to, printing the product name in snake case. Covered by golden tests that replay recorded `git` and `gh` calls, so they need no network.

```
cd scripts && python3 -m new_resources v7.45.0
```

Also adds `gh` to the root `mise.toml` (the only dependency the repo toolchain was missing; `python` was already there) and ignores `__pycache__`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -43,3 +43,4 @@ mise.local.toml
 
 # Don't ignore schema.go if it's part of the Go SDK
 !sdk/go/**/schema.go
+**/__pycache__/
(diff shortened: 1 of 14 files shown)


guineveresaenger (approved):
Seems useful! Discoverability might be an issue, is there anywhere else we have where we could point to this script?

LGTM.

Issue #3944, opened 2026-08-21
GCP v10.0: Create the migration guide


pose:
Translate the [upstream v8 upgrade guide](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/website/docs/guides/version_8_upgrade.html.markdown) into Pulumi naming and add every Pulumi-side break on top. Shape and tone: the [v9 guide](https://www.pulumi.com/registry/packages/gcp/how-to-guides/9-0-migration/).

Author once as `docs/version-10-update.md` in the provider repo, then hand-copy to `pulumi/registry` as `themes/default/content/registry/packages/gcp/how-to-guides/10-0-migration.md`. That copy is **manual** — `resourcedocsgen` only fetches `_index.md` and `installation-configuration.md`. Precedents: [pulumi/registry#8694](https://github.com/pulumi/registry/pull/8694) (v9) and [pulumi/registry#5403](https://github.com/pulumi/registry/pull/5403) (v8).

### Acceptance criteria

- [x] `docs/version-10-update.md` lands in the upgrade PR.
- [ ] Every upstream breaking change is translated or explicitly marked not applicable to Pulumi.
- [ ] Every Pulumi-side break, including patch removals, has the required user action.
- [ ] Breaks needing no user action are marked as such, as the v9 guide does.
- [ ] Links the upstream guide, back-links the v9 guide.
- [ ] `pulumi/registry` PR merged, page renders at `/registry/packages/gcp/how-to-guides/10-0-migration/`.

Issue #3939, opened 2026-08-21
GCP v10.0: Create the major upgrade branch


pose:
Create the long-lived branch that integrates upstream v8.0.0 and that prereleases are cut from.

Both previous majors tagged prereleases on an unmerged PR branch (#2355 for v8, #3355 for v9), which leaves the branch outside ci-mgmt's reach so it drifts from current tooling for the whole window. The preferred approach is the multi-branch ci-mgmt tracking described in pulumi/home#4889. Coordinate with that issue rather than duplicating it.

### Acceptance criteria

- [x] Branch exists, named per the major version release playbook.
- [x] `.ci-mgmt.yaml` sets `major-version: 10` and `make ci-mgmt` has been run.
- [x] Go module paths bumped to `/v10` in `provider/go.mod` and `sdk/go.mod`.
- [x] The branch is tracked by ci-mgmt per pulumi/home#4889, or this issue records why not.
- [x] CI green, or every failure triaged and linked.
- [x] Patch removals land as separate commits. No removals were required: all 8 patches rebased cleanly onto upstream v8.0.0. Whether any are now redundant is tracked separately under #3936.

pose:
`upgrade-to-v10-major` is up, cut from `master` at `eff0ef73c5`, with `.ci-mgmt.yaml` at `major-version: 10`, `make ci-mgmt` applied, and `provider/go.mod` and `sdk/go.mod` on `/v10`. Upstream is at `terraform-provider-google-beta` v8.0.0 and all SDKs are regenerated.

No patches were removed. All 8 rebased onto v8.0.0 cleanly, and the checked-in patch files differ only by hunk offsets with no content changes, so none were absorbed or invalidated upstream. Flagging that explicitly since the acceptance criteria anticipated removals.

What did need removing were 5 stale resource mappings in `provider/resources.go` for resources v8 deleted (`google_ml_engine_model`, `google_iap_brand`, `google_iap_client`, `google_notebooks_environment`, `google_notebooks_instance`). Schema went 1474 to 1461 resources and 506 to 502 functions. `docs/version-10-update.md` documents the user-facing impact in #3966.

[...]

PR #3966, opened 2026-08-27
Add the v10 upgrade guide

pose:
Documents the breaking changes users will hit moving from 9.x to 10.0.0, following the structure of the v9 guide. Content is derived from the upstream `terraform-provider-google-beta` v8.0.0 upgrade guide, with resource and field names translated to Pulumi tokens.

Covers 10 removed resources (BeyondCorp app connections, IAP OAuth, ML Engine, Notebooks, Vertex AI schedules), 6 removed functions, and 19 resources with field-level changes.

Part of #3939
Part of #3944


pose:
Worth covering in the guide: previewing `gcp-py-instance-nginx` from `pulumi/examples` against the v10 alpha emits a deprecation warning that users pinning to v10 will start seeing.

    warning: verification warning: property "metadata" is deprecated: The option to deploy a container
    during VM creation using the container startup agent is deprecated. Use alternative services to run
    containers on your VMs.

It fires on `compute.Instance` when `metadata` carries the container declaration that the VM startup agent consumes, in this shape:

    instance = compute.Instance(
        "poc",
        machine_type="f1-micro",
        metadata={"gce-container-declaration": container_declaration},
    )

It is a warning, not an error, and nothing breaks at v10. The mitigation is not a property rename, so a find-and-replace will not do: Google is retiring the container-on-VM startup agent itself, and the guidance is to run the container through a service built for it, Cloud Run or GKE, or to install a container runtime in a startup script and start the container yourself.

Suggest a short "deprecated, not removed" note so people hitting the warning know it is expected at v10 and what the actual path off it is. Reference: https://cloud.google.com/compute/docs/containers/migrate-containers

